PR #5, opened 2025-07-21
Building, CLI tool & Embedding Server Fixed

andylizf:
1. Building time shorten to less than 40s
2. CLI tool, use `leann <cmd>`
3. Embedding model manager fixed. Do not do reuse anymore. Find a free port and then use it.


yichuan-w:
@andylizf plz add cli info into main

PR #8, opened 2025-07-28
fix: auto-detect normalized embeddings and use cosine distance

andylizf:
- Add automatic detection for normalized embedding models (OpenAI, Voyage AI, Cohere)
- Automatically set distance_metric='cosine' for normalized embeddings
- Add warnings when using non-optimal distance metrics
- Implement manual L2 normalization in HNSW backend (custom Faiss build lacks normalize_L2)
- Fix DiskANN zmq_port compatibility with lazy loading strategy
- Add documentation for normalized embeddings feature

This fixes the low accuracy issue when using OpenAI text-embedding-3-small model with default MIPS metric.

### DIFF
--- a/docs/normalized_embeddings.md
+++ b/docs/normalized_embeddings.md
@@ -1,0 +1,75 @@
+# Normalized Embeddings Support in LEANN
+
+LEANN now automatically detects normalized embedding models and sets the appropriate distance metric for optimal performance.
+
+## What are Normalized Embeddings?
+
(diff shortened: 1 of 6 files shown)


yichuan-w:
OpenAI embedding seems to work now! LGTB

yichuan-w:
@andylizf 
Still some bug when I run:
python examples/google_history_reader_leann.py --max-entries -1 --index-dir wo_recompute_openai_new --query "Tell me my browser history about some coference i often visit"

PR #16, opened 2025-08-04
feat: Add smart memory configuration for DiskANN

andylizf:
## Summary
Auto-configure DiskANN memory parameters based on data size and system specs.

## Changes
- **search_memory_maximum**: 1/10 of embedding size (controls PQ compression)
- **build_memory_maximum**: 50% of available RAM (controls sharding)
- Automatic fallback to defaults if manually specified

## Benefits  
- Optimal performance for different dataset sizes
- Better memory utilization on various hardware
- No manual tuning required

## Testing
✅ Tested with small dataset (8 vectors, 768 dims)
✅ No crashes, successful index building
✅ Memory allocation as expected


yichuan-w:
LGTB

PR #18, opened 2025-08-05
docs: Update co-contributors with GitHub usernames

andylizf:


README.md
```diff
@@ -574,8 +574,9 @@ MIT License - see [LICENSE](LICENSE) for details.
 
 ## 🙏 Acknowledgments
 
-This work is done at [**Berkeley Sky Computing Lab**](https://sky.cs.berkeley.edu/).
+Co-contributors: [Yichuan Wang](https://github.com/yichuan-w) & [Zhifei Li](https://github.com/andylizf).
```
yichuan-w: change to contributors?
because we mey add more in the future?
Also, set my homepage rather than github plz

yichuan-w: @andylizf

yichuan-w: Co-contributors: -> Contributors:
as we may have more ppl (potentially?)in the future

yichuan-w: to show that we would love to collabrate with more ppl

PR #21, opened 2025-08-08
Propagate hosts argument for ollama through chat.py

joshuashaffer:
Propagate the llm-hosts argument for ollama through chat.py so it can use ollama on a different machine.


andylizf (approved):
Awesome, thanks @joshuashaffer so much for the contribution and for fixing this! LGTM.

yichuan-w:
Nice fix, thanks

PR #29, opened 2025-08-11
feat(core,diskann): robust embedding server (no-hang) + DiskANN fast mode (graph partition)

andylizf:
This pull request consolidates the robust embedding server lifecycle and adds a fast mode for DiskANN graph partitioning.

***

#### Embedding Server Hardening (`HNSW` & `DiskANN`)
- A **per-instance manager** is now used, eliminating cross-process reuse and scanning.
- Default shutdown is handled by `atexit` and `weakref.finalize`.
- ZMQ options `SNDTIMEO=1s` and `LINGER=0` have been set to prevent send blocking during shutdown.
- The server now provides shape-correct responses and tolerates nested `[[ids]]`.

#### DiskANN Fast Mode (Graph Partition)
- **Automatic partitioning** now occurs during the build process when `is_recompute=True`.
- The system will **auto-detect partition files** on load and set the `partition_prefix`.
- Implemented **safe cleanup** of large `_disk.index` artifacts after partitioning is complete.

***

### ⚠️ Breaking Changes

- **Cross-process server reuse has been removed**. Servers are now always started fresh for each process.

***

### ➡️ Migration Guide

- It is now preferred to call `cleanup()` explicitly in long-lived applications. The `atexit` hook will handle default shutdown cases.


yichuan-w:
@andylizf , lets make the change back-compatible, thanks!

PR #45, opened 2025-08-14
Docs/Core: Low-Resource Setups, SkyPilot Option, and No-Recompute

### LINKED ISSUE #43: What are suggested system requirements for LEANN?
I don't have a GPU on my linux VM and it is quite slow to index and search, even with small models. Curious if there are general minimum system requirements you suggest for LEANN to work well?

Also, I can't figure out how to turn off recompute on search. Is this functionality available today?

andylizf:
Fixes #43.

---

### **Key Changes**

* **Documentation:** Clarified and added support for low-resource setups, including:
    * OpenAI embeddings
    * SkyPilot remote build
    * `--no-recompute` usage
* **CLI:** Added more intuitive flags:
    * `--no-recompute` (for `build`/`search`/`ask`)
* **HNSW (Hierarchical Navigable Small World):** If `is_recompute=False` and `is_compact=True`, the process will now automatically force a non-compact build and issue a warning for compatibility.
* **Index Portability:**
    * The index metadata now keeps relative path entries.
    * Added optional `path_relative` and `index_path_relative` hints for better portability.
* **Reader:** The `PassageManager` can now resolve absolute paths that don't exist by falling back to relative path hints or sibling filenames, which is a non-destructive operation.
* **Embedding Server:** Path resolution is now delegated to the `PassageManager` via `metadata_file_path` to ensure consistent behavior across the system.
* **SkyPilot Integration:**
    * Fixed environment variable configurations.
    * Added tunable parameters.
    * Ensured C++ runtime libraries are installed.
    * Made sure the remote index directory exists before use.
    * Documented the `rsync` path for transferring data.

---

### **Verification**

* Successfully built an index remotely on AWS using SkyPilot.
* Synced the index back to a local machine using `rsync`.
* Confirmed that local search works correctly (with recompute).
* Confirmed that the `--no-recompute` flag works as expected when building a non-compact HNSW index.


packages/leann-core/src/leann/cli.py
```diff
@@ -161,10 +180,17 @@ def create_parser(self) -> argparse.ArgumentParser:
         ask_parser.add_argument("--beam-width", type=int, default=1)
         ask_parser.add_argument("--prune-ratio", type=float, default=0.0)
         ask_parser.add_argument(
-            "--recompute-embeddings",
+            "--no-recompute",
+            dest="recompute_embeddings",
```
yichuan-w: the design of these two args looks not that good

yichuan-w:
Hi @andylizf , I have another two comments 1. do we really skip delete the embedding operation during no-recompute
2. we need a assert error in the search if we provide recompute in the build but requires no recompute in search

yichuan-w:
also a quick test is that if you found the storage is larger than recompute and it is faster then it is correct, also seems now we do not support diskann no recompute, it is fine but we need to deal with them in parser @andylizf

yichuan-w:
@andylizf also, it would be nice, if you can have a robust add API, I assue it can work well with no-recompute but not that sure about w recompute, maybe we should change the backend(I guess the answer is yesy as we should pass the args in the add api, we can do that in next/future PR)

README.md
```diff
@@ -512,7 +514,8 @@ leann search INDEX_NAME QUERY [OPTIONS]
 Options:
   --top-k N                   Number of results (default: 5)
   --complexity N              Search complexity (default: 64)
-  --recompute-embeddings      Use recomputation for highest accuracy
+  --recompute                 Use recomputation for highest accuracy
+  --no-recompute              Disable recomputation (requires non-compact HNSW index)
```
yichuan-w: is that HNSW specific? I guess you are adding diskann as well, so let's fix this, also better to show this in --h

README.md
```diff
@@ -512,7 +514,8 @@ leann search INDEX_NAME QUERY [OPTIONS]
 Options:
   --top-k N                   Number of results (default: 5)
   --complexity N              Search complexity (default: 64)
-  --recompute-embeddings      Use recomputation for highest accuracy
+  --recompute                 Use recomputation for highest accuracy
```
yichuan-w: remove the highest accuracy cuz basically it is the same accuracy

benchmarks/benchmark_no_recompute.py
```diff
@@ -0,0 +1,148 @@
...
+    size_diskann_nr = _size_for(diskann_nr)
+
+    print("\nBenchmark results (DiskANN):")
+    print(f"  build(recompute=True, partition): size={size_diskann_r / 1024 / 1024:.1f}MB")
+    print(f"  build(recompute=False):          size={size_diskann_nr / 1024 / 1024:.1f}MB")
+    print(f"  search recompute=True (final rerank): {t_diskann_r:.3f}s")
+    print(f"  search recompute=False (PQ only):     {t_diskann_nr:.3f}s")
+
+
+if __name__ == "__main__":
```
yichuan-w, line 147: Can you put a reference result here? maybe in the beginning

docs/configuration-guide.md
```diff
@@ -273,24 +267,114 @@ Every configuration choice involves trade-offs:
...
+  --embedding-mode openai \
+  --embedding-model text-embedding-3-small
+
+# Search with OpenAI embeddings (recompute at query time)
+leann search my-index "your query" \
+  --recompute
+```
+
+### 2) Run remote builds with SkyPilot (cloud GPU)
+
```
yichuan-w, line 292: fold this, no one will use, ppl dont have GPU will not have remote GPU

yichuan-w:
Great, I have left some comments and please fix these @andylizf, generally, it is good

docs/configuration-guide.md
andylizf: this config file should be comprehensive enough. and it's docs but not readme. we should not fold anything in this docs

PR #58, opened 2025-08-16
Add AST-aware code chunking for better code understanding

gabriel-dehan:
Introduces intelligent AST-aware code chunking to LEANN using [astchunk](https://github.com/yilinjz/astchunk), improving code understanding in RAG applications. Instead of arbitrarily splitting code at character boundaries, we now have the option to preserve semantic structure by chunking at function, class, and method boundaries.

This enhancement should make LEANN significantly better for code-related RAG applications while maintaining full backward compatibility. 

I am also planning on adding `ruby` support to astchunk and have opened a [pull request](https://github.com/yilinjz/astchunk/pull/5) there.

### AST-Aware Code Chunking
- **Smart boundary detection**: Preserves functions, classes, and methods as complete units
- **Multi-language support**: Python, Java, C#, TypeScript out of the box (Ruby soon #PR_LINK)
- **Graceful fallback**: Falls back to traditional chunking for unsupported languages or parsing errors
- **Rich metadata**: Each chunk includes line numbers, file paths, and AST node information

### 📱 New application
- **Code RAG app** (`apps/code_rag.py`): Specialized for code repositories with optimized parameters
- **Enhanced document RAG**: New `--enable-code-chunking` flag for mixed content

### 🛠️ CLI Integration
- Global `--use-ast-chunking` flag for the main CLI
- Configurable chunk sizes and overlap for both code and text
- Proper error handling and user feedback

## How this helps

**Before**: Code was chunked arbitrarily, often splitting functions mid-way
```python
# Traditional chunking might split this:
def important_function():
    """This docstring explains the function."""
    # Implementation here...
--- CHUNK BOUNDARY ---
    return result
```

**After**: AST chunking preserves semantic boundaries
```python
# AST chunking keeps this together:
def important_function():
    """This docstring explains the function."""
    # Complete implementation
    return result
```

This should result in a much better code understanding and search quality in RAG applications.

## Examples

```bash
# Enable AST chunking for mixed content
python -m apps.document_rag --enable-code-chunking --data-dir ./my_project

# Use the specialized code RAG
python -m apps.code_rag --repo-dir ./codebase --query "How does authentication work?"

# Global CLI with AST support
leann build my-index --docs ./src --use-ast-chunking
```

## Files Changed

- `apps/chunking_utils.py` - New AST-aware chunking utilities
- `apps/code_rag.py` - Specialized code RAG application  
- `apps/base_rag_example.py` - Enhanced with AST chunking arguments
- `apps/document_rag.py` - Added code chunking flag
- `packages/leann-core/src/leann/cli.py` - CLI integration
- `pyproject.toml` - Added astchunk dependencies
- `tests/test_astchunk_integration.py` - Comprehensive test suite
- `README.md` & `docs/features.md` - Updated documentation

## Dependencies

Added optional dependencies for AST parsing:
- `astchunk>=0.1.0` - Core AST chunking library
- `tree-sitter>=0.20.0` + language parsers - AST parsing backend

### DIFF
--- a/ASTCHUNK_INTEGRATION.md
+++ b/ASTCHUNK_INTEGRATION.md
@@ -1,0 +1,228 @@
+# AST-Aware Code Chunking Integration
+
+## Overview
+
+This document describes the integration of **astchunk** library into LEANN for AST-aware code chunking. This enhancement provides better semantic understanding of code structure compared to traditional text-based chunking.
+
+## 🚀 Features Added
+
+### 1. Enhanced Chunking Utilities (`apps/chunking_utils.py`)
+- **AST-aware chunking** for Python, Java, C#, TypeScript files
+- **Automatic language detection** from file extensions  
+- **Fallback mechanisms** to traditional chunking when AST fails
+- **Configurable chunk sizes** and overlap parameters
+- **Graceful error handling** for unsupported languages
+
+### 2. Updated Document RAG (`apps/document_rag.py`)
+- New `--enable-code-chunking` flag for AST-aware processing
+- Backward compatibility with existing workflows
+- Automatic detection and processing of code vs text files
+
+### 3. Specialized Code RAG Application (`apps/code_rag.py`)
+- Dedicated application for code repository indexing
+- Optimized parameters for code understanding
+- Automatic exclusion of build directories and cache files
+- Support for common code file extensions
+
+### 4. Enhanced CLI Support (`packages/leann-core/src/leann/cli.py`)
+- New AST chunking command-line arguments:
+  - `--use-ast-chunking`: Enable AST-aware chunking
+  - `--ast-chunk-size`: Set AST chunk size (default: 768)
+  - `--ast-chunk-overlap`: Set AST chunk overlap (default: 96)
+  - `--ast-fallback-traditional`: Enable fallback mode
+
+### 5. Comprehensive Test Suite (`tests/test_astchunk_integration.py`)
+- Unit tests for chunking functions
+- Integration tests with document RAG
+- Error handling and edge case testing
+- Mock objects for testing without dependencies
+
+## 🛠️ Installation
+
+### Dependencies Added to `pyproject.toml`
+```toml
+# AST-aware code chunking dependencies
+"astchunk>=0.1.0",
+"tree-sitter>=0.20.0", 
+"tree-sitter-python>=0.20.0",
+"tree-sitter-java>=0.20.0",
+"tree-sitter-c-sharp>=0.20.0",
+"tree-sitter-typescript>=0.20.0",
+```
+
+### Install Dependencies
+```bash
+# Install LEANN with AST chunking support
+uv pip install -e "."
+
+# Or install dependencies manually
+uv pip install astchunk tree-sitter tree-sitter-python tree-sitter-java tree-sitter-c-sharp tree-sitter-typescript
+```
+
+## 📖 Usage Examples
+
+### 1. Document RAG with Code Chunking
+```bash
+# Enable AST chunking for mixed content
+python -m apps.document_rag --enable-code-chunking --data-dir ./my_project
+
+# Query code and documentation together
+python -m apps.document_rag --enable-code-chunking --query "How does the authentication system work?"
+```
+
+### 2. Specialized Code RAG
+```bash
+# Index an entire code repository  
+python -m apps.code_rag --repo-dir ./my_codebase
+
+# Index with custom settings
+python -m apps.code_rag --repo-dir ./src --ast-chunk-size 1024 --query "Show me the database connection logic"
+
+# Include specific file types
+python -m apps.code_rag --include-extensions .py .js .ts --query "Find all API endpoints"
+```
+
+### 3. Global CLI with AST Support
+```bash
+# Build index with AST chunking
+leann build my-code-index --docs ./src --use-ast-chunking --ast-chunk-size 512
+
+# Search with traditional CLI
+leann search my-code-index "database connection"
+leann ask my-code-index --interactive
+```
+
+### 4. Advanced Configuration
+```bash
+# Fine-tune AST chunking parameters
+python -m apps.code_rag \
+    --repo-dir ./large_project \
+    --ast-chunk-size 1024 \
+    --ast-chunk-overlap 128 \
+    --max-file-size 2000000 \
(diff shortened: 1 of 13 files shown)


yichuan-w:
Thanks for your contribution!! It is a very important feature, we will review and merge asap.

apps/code_rag.py
```diff
@@ -0,0 +1,208 @@
...
+
+import sys
+from pathlib import Path
+
+# Add parent directory to path for imports
+sys.path.insert(0, str(Path(__file__).parent))
+
+from base_rag_example import BaseRAGExample, create_text_chunks
+from llama_index.core import SimpleDirectoryReader
+
```
andylizf, line 16:
```suggestion
from .utils import CODE_EXTENSIONS
```

yichuan-w:
We recently came across [astchunk](https://github.com/yilinjz/astchunk), the pkg you are using and its accompanying [paper](https://arxiv.org/html/2506.15655v1), too. Have you had a chance to evaluate it or compare it against naive chunking? I’m not entirely convinced about the extent of its performance gains, but it could be worth discussing further.

ASTCHUNK_INTEGRATION.md
andylizf, line 1: Let's move it to `docs/`?

andylizf: Seems this file is a little bit overwhelming. Can we simplify it and make this docs the best practices for code raging, like providing some examples here. Also we can cite this docs in our claude code integration `leann_mcp/README.md`

packages/leann-core/src/leann/cli.py
```diff
@@ -726,18 +750,52 @@ def file_filter(
...
+                    ast_chunk_overlap=getattr(args, 'ast_chunk_overlap', 96),
+                    code_file_extensions=None,  # Use defaults
+                    ast_fallback_traditional=getattr(args, 'ast_fallback_traditional', True)
+                )
+                
+            except ImportError as e:
+                print(f"⚠️  AST chunking not available ({e}), falling back to traditional chunking")
+                use_ast = False
+        
+        if not use_ast:
```
andylizf, line 1197:
```suggestion
        else:
```

packages/leann-core/src/leann/cli.py
```diff
@@ -726,18 +750,52 @@ def file_filter(
...
-                all_texts.append(node.get_content())
+        
+        # Check if AST chunking is requested
+        use_ast = getattr(args, 'use_ast_chunking', False)
+        
+        if use_ast:
+            print("🧠 Using AST-aware chunking for code files")
+            try:
+                # Import enhanced chunking utilities
+                import sys
```
andylizf: lets move it up.

packages/leann-core/src/leann/cli.py
```diff
@@ -421,6 +444,7 @@ def load_documents(
         docs_paths: Union[str, list],
         custom_file_types: Union[str, None] = None,
         include_hidden: bool = False,
+        args=None
```
andylizf: lets have type hints here
```suggestion
        args: Optional[Dict[str, Any]] = None
```

andylizf (changes requested):
Mostly LGTM! Thanks for your hard work @gabriel-dehan ! Left some discussions.

andylizf:
Also, in the future, we may want to refactor our `document_rag` to make it the default implementation of leann cli, which will reduce a lot of redundant code.

yichuan-w:
https://github.com/yichuan-w/LEANN/blob/da6540decd08bd69b965a85aa99d84dc0e04ba08/docs/CONTRIBUTING.md#setup-pre-commit

@gabriel-dehan it is better to have pre-commit, thanks!

gabriel-dehan:
> We recently came across [astchunk](https://github.com/yilinjz/astchunk), the pkg you are using and its accompanying [paper](https://arxiv.org/html/2506.15655v1), too. Have you had a chance to evaluate it or compare it against naive chunking? I’m not entirely convinced about the extent of its performance gains, but it could be worth discussing further.

I've been asking myself the same thing. I've got no idea on how to measure that properly so I used Claude Code to generate a quick and dirty evaluation framework on small and big codebases, it does a lot of things and isn't very clean but it clones big repositories like VSCode / Node / Jango and performs complex queries on it amongst other things.

The results are here: https://github.com/gabriel-dehan/LEANN/pull/2 
I am not saying we should trust them 100%, I've skimmed over the evaluation code and it seems coherent overall but there is way too much stuff there and a lot I don't have the time to understand. But I feel it gives an idea, and the results are not that far from what I'd expect (except for the memory and speed difference, I am a bit skeptical about this)

I'll check your other comments and implement the fixes.

yichuan-w:
Yeah, thanks for the benchmark. I briefly looked at that, and I am confused about the efficiency and memory-saving gain. Maybe I can investigate that later, but at least we can first merge this PR and make it an option, thanks a lot!! Both this PR and the benchmark are great.

Let's fix this PR and merge it first

apps/code_rag.py
gabriel-dehan: This wouldn't work as we are not in the same directory no? We've got an __init__.py in chunking/  so `chunking` should be correct (or `chunking.utils`, but not `.utils`)
I've left it as is

packages/leann-core/src/leann/cli.py
gabriel-dehan: Actually no because we want to fallback if there is an error (see line right above, we set `use_ast = False`) so the next if needs to be separated and not an alternative branch of the previous condition!

ASTCHUNK_INTEGRATION.md
gabriel-dehan: I simplified it and moved it to `docs/`!

gabriel-dehan:
Rebased on main, made the requested changes (for the most part) and installed the pre-commit hook and fixed a bunch of violations. Also commited the uv.lock so the PR seems like a lot more changes than what it actually has 🤭

yichuan-w:
I guess still some lint error? @gabriel-dehan and thanks for fixing!

gabriel-dehan:
Yeah I'll try to look at it tonight!

yichuan-w:
LGTM, let's wait for CI and we will add you to our ack!

yichuan-w:
Looks good, we can merge first and make it default false, if it works well, we can make it default true!

PR #77, opened 2025-08-26
Fix gitignore filtering to properly exclude ignored files

sc0rch:
## Summary
This PR fixes an issue where the CLI was not properly filtering files according to `.gitignore` rules, causing thousands of unnecessary files to be indexed.

## Problem
When indexing a directory with a `.gitignore` file, the CLI would still process files from ignored directories like `venv/`, `node_modules/`, `.mypy_cache/`, etc. 

For example, in a Python project with a virtual environment, the tool would report loading 2461 documents instead of the expected ~55 actual source files.

## Root Cause
The `gitignore_parser` library requires **absolute paths** for correct pattern matching, but the code was passing relative paths. This caused the matcher to fail silently and include files that should have been ignored.

## Solution
Updated three locations in `cli.py` where paths are checked against gitignore patterns:
1. PDF file filtering - now passes absolute paths directly to the matcher
2. `file_filter` function - simplified to use absolute paths
3. `_should_exclude_file` method - uses `.as_posix()` for consistent path formatting

## Testing
Tested locally with a Python project containing:
- `venv/` directory with 5982 files
- `.mypy_cache/` with 1157 files  
- Other ignored directories

**Before fix:** Loaded 2461 documents
**After fix:** Loaded 55 documents (correctly excluding venv and cache directories)

## Impact
This fix significantly reduces:
- Indexing time (no need to process thousands of irrelevant files)
- Storage requirements (fewer chunks to store)
- Memory usage during indexing
- Overall better user experience matching expected gitignore behavior

Fixes the issue where `.gitignore` rules are mentioned in the README but not working correctly in practice.


sc0rch:
I'm not an expect in Python, just saying, basically I asked Claude to fix leann, it was parsing 6k files in a small Python repository.

If you see anything useful in this PR, please fix, thanks!

If Claude was wrong about the code, sorry about that, in this case just close the PR.

yichuan-w:
Thanks for your PR We will take a look at this.

btw, have you tried 
leann build my-repo --docs $(git ls-files) --embedding-mode sentence-transformers --embedding-model all-MiniLM-L6-v2 --backend hnsw
This method should work to ignore stuff in .gitignore

yichuan-w:
close this first, cuz solved by #84

PR #84, opened 2025-09-01
[CLI] More robust leann list and leann build

yichuan-w:


yichuan-w:
For the previous commit.

We previously used resolved (symlink-followed) absolute paths to check if PDFs were inside the --docs directory, which misclassified symlinked files as “outside scope.” Now we check directory scope using the original (non-resolved) path and still use absolute paths for .gitignore matching, eliminating the false “outside” skips.

Issue #86, opened 2025-09-04
Add grep in raw data


yichuan-w:
This should be an easy task to implement BM25 and only use grep or high perf grep method, ultimately we want agent to determine whether semantic search or grep, although claude code can do that lol

R3DK3LL:
Hi! I'd like to work on this issue, new to it all but keen to help if I can, look forward to hearing more about the scope, thanks.

yichuan-w:
@R3DK3LL  basically, this one is very easy

Basically we will have my-docs-test/documents.leann.passages.jsonl to store the raw text
The first solution is you can build BM25 index on this, but I guess it is something heavy weight which I do not like(both building time/storage/search time)
The second one is to run grep on this file and then may use some method to extract the line(I love this)
like

> {"id": "0", "text": "DiskGNN: Bridging I/O Efficiency and Model Accuracy\nfor Out-of-Core GNN Training\nRenjie Liu*, Yichuan Wang*, Xiao Yan, Haitian Jiang, Zhenkun Cai, \nMinjie Wang, Bo Tang, Jinyang Li (* indicates equal contributions)\nOut-of-core GNN Training\nOffline Sampling–based Training\n-\nIn resource-constrained environments, large graphs (topology \n& features) can not be loaded from disk to memory.\n-\nA training system is needed that can fetch data directly from disk \nwhile minimizing I/O overhead.\nDisk\nCPU\n?\nB\nC\nD\nG\nH\nF\nE\nA\nK\nM\nGPU\n?\n?\nChallenge: I/O Efficiency & Model Acc\n-\nIssues of  existing systems:\n-\nHigh I/O volume due to read amplification\n-\nDegrade GNN model accuracy due to loss of edges\nExisting systems either have poor I/O efficiency or degrade \nthe model accuracy.", "metadata": {}}
 

Then we can return the result, for the API design, we may want to pass special args to LEANNsearcher to perform grep style search

Let me know if you need more hints @R3DK3LL

R3DK3LL:
Hey @yichuan-w ,
I've got some new changes for the grep search looking over the code again. It's more reliable and faster. Just checking if you need me to open a new pull request or add the changes to my existing branch.
Thanks

yichuan-w:
Great, I will review that, let's discuss in that PR, thx!

yichuan-w:
this should be impl by #87

PR #87, opened 2025-09-04
Add grep search functionality - Issue #86

R3DK3LL:
Implements grep-based search functionality as requested in issue #86.

**Changes:**
- Added `use_grep` parameter to `LeannSearcher.search()` method
- Implemented grep search on `.jsonl` files containing raw passages  
- Added fallback Python regex search if grep unavailable
- Returns results in same `SearchResult` format as semantic search

**Usage:**
```python
searcher = LeannSearcher("path/to/index")
results = searcher.search("query text", use_grep=True)

### DIFF
--- a/packages/leann-core/src/leann/api.py
+++ b/packages/leann-core/src/leann/api.py
@@ -6,6 +6,8 @@ with the correct, original embedding logic from the user's reference code.
 import json
 import logging
 import pickle
+import subprocess
+import re
 import time
 import warnings
 from dataclasses import dataclass, field
@@ -653,6 +655,7 @@ class LeannSearcher:
         expected_zmq_port: int = 5557,
         metadata_filters: Optional[dict[str, dict[str, Union[str, int, float, bool, list]]]] = None,
         batch_size: int = 0,
+        use_grep: bool = False,
         **kwargs,
     ) -> list[SearchResult]:
         """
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by yichuan-w** [comment]
@R3DK3LL, please fix the lint error first, see https://github.com/yichuan-w/LEANN/blob/main/docs/CONTRIBUTING.md#-pre-commit-hooks
Thanks a lot!

File: packages/leann-core/src/leann/api.py [hunk 1]
```diff
@@ -795,9 +802,94 @@ def search(
...
+                        )
+                    )
+                except json.JSONDecodeError:
+                    continue
+
+            matches.sort(key=lambda x: x.score, reverse=True)
+            return matches[:top_k]
+
+        except FileNotFoundError:
+            return self._python_regex_search(query, top_k)
```
yichuan-w: Is there indeed some case that we cannot grep but we can regex?

yichuan-w:
@R3DK3LL I think this is an amazing PR can you put some examples like this https://github.com/yichuan-w/LEANN/blob/main/examples/spoiler_free_book_rag.py ,
Also, you can make it visible in some place like doc and readme, so people can know how to use that, really a nice feature, thanks a lot!!!

After the PR is merged, I will add you to a **contributor** in the readme

packages/leann-core/src/leann/api.py
R3DK3LL: Good point, the fallback was for sys where grep isn't installed, Py regex would handle the same cases. I can remove the fallback if you prefer an explicit fail-fast error when grep isn't available, let me know.

yichuan-w: I guess we can simply throw error that the Unix does not have free rather than regex match?

yichuan-w:
@R3DK3LL  can we remove the example of grep in our main example and mention it later? Like in feature section and link to your code? Because I want to keep the example simple enough, thanks

yichuan-w:
Great, lets change the readme as my previous comment and then we can merge , thanks @R3DK3LL

yichuan-w:
LGTM!! Thanks! I will merge after passing the CI.

PR #98, opened 2025-09-14
[FEAT] Dynamic add

yichuan-w:
In this PR, we will add LEANN-based dynamic add

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,4 @@ CLAUDE.local.md
 .claude/*.local.*
 .claude/local/*
 benchmarks/data/
+test_add/*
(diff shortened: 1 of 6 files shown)


yichuan-w:
python examples/dynamic_add_leann_no_recompute.py --build-base \
    --recompute-build --ef-construction 200 \
    --base-dir /Users/yichuan/Desktop/code/LEANN/leann/data \
    --index-dir ./test_doc_files --index-name documents.leann
python examples/dynamic_add_leann_no_recompute.py --add-incremental \
    --recompute-add --ef-construction 32 \
    --add-dir /Users/yichuan/Desktop/code/LEANN/leann/test_add \
    --index-dir ./test_doc_files --index-name documents.leann
    
    
  And we usually fail at the second element to add in the second command, @andylizf , can you take a look at this.
  Basically it is caused by omp if we use lldb to debug

PR #75, opened 2025-08-19
Metadata filtering feature

gabriel-dehan:
Hey guys,

I've been working on a feature that I need for one of my personal projects. I am building a small spoiler-free RAG system to retrieve information from book chapters without spoiling yourself (let's say you are taking up again a book you started reading a few years ago and don't want to read the book again, you can ask questions about characters, etc, up to the chapter you are currently reading).

For this I needed to add a way to filter search results by chapters amongst other things and I thought it could prove to be a useful generic feature to have integrated in LEANN. So this PR lets you filter search results based on custom metadata fields you attach to your documents during chunking / indexing.

## What this enables

- Code search filtering: Filter by file extension, lines of code, date modified
- Document management: Filter by document type, creation date, author
- Content filtering: Filter by tags, categories, chapter number, etc.

## How it works

```python
# Add metadata when building your index
builder.add_text(
    "def authenticate_user(token): ...",
    metadata={"file_extension": ".py", "lines_of_code": 25}
)

# Filter during search
results = searcher.search(
    query="authentication function",
    metadata_filters={
        "file_extension": {"==": ".py"},
        "lines_of_code": {"<": 100}
    }
)
```

Filtering is implemented as a post-search step, so it doesn't touch the backend interfaces, it might be a bit more limited this way but with the way the system is implemented at the time, it seems nigh impossible to do it another way.

## What's included

- Comprehensive filter operators: ==, !=, <, <=, >, >=, in, not_in, contains, starts_with, ends_with, is_true, is_false
- Multiple field support: Filters combine using AND logic
- Documentation: Usage guide with examples
- Working example: Spoiler-free book RAG demo

Cheers!

PR #103, opened 2025-09-19
Implement FileSystem wide semantic file search engine with temporal awareness using LEANN. 

Kishlay-notabot:
This example code is a collaborative effort between the monkeSearch project and the developers of LEANN to bring semantic file search capabilities to PCs even without a GPU. Draft PR opened for optimization suggestions and further additions. 

This is a project of its own developed by me named [monkeSearch](https://github.com/monkesearch/monkeSearch/). There are other implementation techniques and ongoing development happening on the repo to make it available as a general use tool with GUI/TUI etc.

### DIFF
--- a/apps/semantic_file_search/leann-plus-temporal-search.py
+++ b/apps/semantic_file_search/leann-plus-temporal-search.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+from pathlib import Path
+from leann import LeannSearcher
+import sys
+import re
+from datetime import datetime, timedelta
+
+INDEX_PATH = str(Path("./").resolve() / "demo.leann")
+
+class TimeParser:
+    def __init__(self):
+        # Main pattern: captures optional fuzzy modifier, number, unit, and optional "ago"
+        self.pattern = r'(?:(around|about|roughly|approximately)\s+)?(\d+)\s+(hour|day|week|month|year)s?(?:\s+ago)?'
+        
+        # Compile for performance
+        self.regex = re.compile(self.pattern, re.IGNORECASE)
+        
+        # Stop words to remove before regex parsing
+        self.stop_words = {
+            'in', 'at', 'of', 'by', 'as', 'me',
+            'the', 'a', 'an', 'and', 'any',
+            'find', 'search', 'list',
+            'ago', 'back',
+            'past', 'earlier', 
+        }
+    
+    def clean_text(self, text):
+        """Remove stop words from text"""
+        words = text.split()
+        cleaned = ' '.join(word for word in words if word.lower() not in self.stop_words)
+        return cleaned
+    
+    def parse(self, text):
+        """Extract all time expressions from text"""
+        # Clean text first
+        cleaned_text = self.clean_text(text)
+        
+        matches = []
+        for match in self.regex.finditer(cleaned_text):
+            fuzzy = match.group(1)  # "around", "about", etc.
+            number = int(match.group(2))
+            unit = match.group(3).lower()
+            
+            matches.append({
+                'full_match': match.group(0),
+                'fuzzy': bool(fuzzy),
+                'number': number,
+                'unit': unit,
+                'range': self.calculate_range(number, unit, bool(fuzzy))
+            })
+        
+        return matches
+    
+    def calculate_range(self, number, unit, is_fuzzy):
+        """Convert to actual datetime range and return ISO format strings"""
+        units = {
+            'hour': timedelta(hours=number),
+            'day': timedelta(days=number),
+            'week': timedelta(weeks=number),
+            'month': timedelta(days=number * 30),
+            'year': timedelta(days=number * 365)
+        }
+        
+        delta = units[unit]
+        now = datetime.now()
+        target = now - delta
+        
+        if is_fuzzy:
+            buffer = delta * 0.2  # 20% buffer for fuzzy
+            start = (target - buffer).isoformat()
+            end = (target + buffer).isoformat()
+        else:
+            start = target.isoformat()
+            end = now.isoformat()
+        
+        return (start, end)
+
+def search_files(query, top_k=15):
+    """Search the index and return results"""
+    # Parse time expressions
+    parser = TimeParser()
+    time_matches = parser.parse(query)
+    
+    # Remove time expressions from query for semantic search
+    clean_query = query
+    if time_matches:
+        for match in time_matches:
+            clean_query = clean_query.replace(match['full_match'], '').strip()
+    
+    # Single query to vector DB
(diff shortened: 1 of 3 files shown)


yichuan-w:
Thanks for your contribution. Please follow our contribution guide to fix the format

yichuan-w:
please follow the format check https://github.com/yichuan-w/LEANN/blob/main/docs/CONTRIBUTING.md#-pre-commit-hooks

yichuan-w:
@andylizf , can you take a look on this

yichuan-w:
@Kishlay-notabot tell us when the PR is ready! And @andylizf will review that, thanks for the promotion on Reddit and welcome to LEANN

yichuan-w:
@andylizf can you help to review this PR?

yichuan-w:
We usually recommend post-filter, https://github.com/yichuan-w/LEANN/pull/75#issuecomment-3204403449, and we have a previous impl here

apps/semantic_file_search/leann-plus-temporal-search.py
```diff
@@ -0,0 +1,168 @@
...
+    """Search the index and return results"""
+    # Parse time expressions
+    parser = TimeParser()
+    time_matches = parser.parse(query)
+
+    # Remove time expressions from query for semantic search
+    clean_query = query
+    if time_matches:
+        for match in time_matches:
+            clean_query = clean_query.replace(match["full_match"], "").strip()
```
andylizf, line 106: After stripping digits you still pass "ago" (or similar stop-words) into the semantic search, so pure time queries return no hits. We needs additional cleanup or a fallback to the original query here.

apps/semantic_file_search/leann_index_builder.py
```diff
@@ -0,0 +1,54 @@
...
+        # Add to builder
+        builder.add_text(embedding_text, metadata=metadata)
+
+        # Show progress
+        progress = (idx + 1) / total_items * 100
+        sys.stdout.write(f"\rProgress: {idx + 1}/{total_items} ({progress:.1f}%)")
+        sys.stdout.flush()
+
+    print("\n\nBuilding index...")
+    builder.build_index(INDEX_PATH)
```
andylizf: An empty Spotlight dump hits `LeannBuilder.build_index` and `raises ValueError("No chunks added.");` guard for zero items and exit gracefully instead.

apps/semantic_file_search/leann-plus-temporal-search.py
```diff
@@ -0,0 +1,168 @@
...
+
+    # Single query to vector DB
+    searcher = LeannSearcher(INDEX_PATH)
+    results = searcher.search(
+        clean_query if clean_query else query, top_k=top_k, recompute_embeddings=False
```
andylizf, line 116: Time filtering runs only after retrieval, so any in-range results outside the initial top-K are lost. We might need to push the range into `metadata_filters` so the backend enforces it.
let's double check how this `metadata_filters` work.

andylizf (changes requested):
Left some discussions.

apps/semantic_file_search/leann-plus-temporal-search.py
Kishlay-notabot: I'm really not sure what to include in a fallback query, vector search can be unpredictable then? am i right?

apps/semantic_file_search/leann-plus-temporal-search.py
Kishlay-notabot: Yes, I don't have much idea on it :/ 
What should we do? I think we should not worry about an empty query containing only a time filter, because we also cannot set a fallback query. How should we approach on these problems? I think we should invalidate the user query after regex check if there's no usable keywords. Might be a solution for this and the first fallback query problem.

apps/semantic_file_search/leann-plus-temporal-search.py
Kishlay-notabot: I've added a simple check, to disable time only searches. Not sure how would we proceed with the default query for the db.

apps/semantic_file_search/leann_index_builder.py
```diff
@@ -0,0 +1,82 @@
...
+    builder = LeannBuilder(backend_name="hnsw", is_recompute=False)
+
+    total_items = len(items)
+    items_added = 0
+    print(f"Processing {total_items} items...")
+
+    for idx, item in enumerate(items):
+        try:
+            # Create embedding text sentence
+            embedding_text = f"{item.get('Name', 'unknown')} located at {item.get('Path', 'unknown')} and size {item.get('Size', 'unknown')} bytes with content type {item.get('ContentType', 'unknown')} and kind {item.get('Kind', 'unknown')}"
```
andylizf, line 30: Only metadata is embedded. not the file content. is this expected?

apps/semantic_file_search/leann-plus-temporal-search.py
```diff
@@ -0,0 +1,173 @@
...
+        start_time, end_time = time_range
+
+        filtered_results = []
+        for result in results:
+            # Access metadata attribute directly (not .get())
+            metadata = result.metadata if hasattr(result, "metadata") else {}
+
+            if metadata:
+                # Check modification date first, fall back to creation date
+                date_str = metadata.get("modification_date") or metadata.get("creation_date")
```
andylizf, line 131: is this `str` type?

andylizf: iiuc we should first convert them to `datetime`

apps/semantic_file_search/spotlight_index_dump.py
```diff
@@ -0,0 +1,260 @@
...
+Spotlight Metadata Dumper for Vector DB
+Extracts only essential metadata for semantic search embeddings
+Output is optimized for vector database storage with minimal fields
+"""
+
+import json
+import sys
+from datetime import datetime
+
+from Foundation import NSDate, NSMetadataQuery, NSPredicate, NSRunLoop
```
andylizf, line 17: should check if it's on macos platform.

apps/semantic_file_search/leann_index_builder.py
Kishlay-notabot: yes, for now. 
File content chunks are a feature I am looking forward to, but currently aiming at optimizing and working on the first prototype with this feature only.

yichuan-w:
@andylizf . is this PR ready? If it is, please merge that

yichuan-w:
@Kishlay-notabot @andylizf lets review this PR again if this is ready let's add some instruction and then merge it to main branch

andylizf (approved):
LGTM

PR #108, opened 2025-09-20
Introducing dynamic index update

andylizf:
Supersedes #98.

## Related Issues

Fixes #

## Checklist


examples/dynamic_update_no_recompute.py
```diff
@@ -0,0 +1,404 @@
...
+
+    print(f"\n{prefix}Updating index with additional passages...")
+    update_index(
+        index_path,
+        start_id=len(initial_paragraphs),
+        paragraphs=update_paragraphs,
+        model_name=model_name,
+        embedding_mode=embedding_mode,
+        is_recompute=is_recompute,
+    )
```
yichuan-w, line 216: We can support both is_recompute=false and true, right?

I guess it's better to test both,
one is to no recompute add(base and update)
one is recompute add(base and update)

andylizf: Yes we are testing both.

https://github.com/yichuan-w/LEANN/blob/6157bebb9e441c05ecd6a8dd0f31df21c9981980/examples/dynamic_update_no_recompute.py#L390-L393

yichuan-w:
LGTM, feel free to merge!

PR #116, opened 2025-09-23
Fix 'leann ask' positional query handling

### LINKED ISSUE #110: Ask example given by leann doesn't work
### What happened?

I was getting set up with leann and was trying the example:

```
  leann ask my-docs "question"                                           # Ask my-docs index
```

Unfortunately trying this doesn't work.  Looks like the question part isn't supported:

```
❯ leann ask agent "where are prompts configured"
usage: leann [-h] {build,search,ask,list,remove} ...
leann: error: unrecognized arguments: where are prompts configured
```

### How to reproduce

```
leann ask agent "where are prompts configured"
```

### Error message

```shell

```

### LEANN Version

0.1.0

### Operating System

macOS

andylizf:
Fixes #110.

## Summary
- allow the ask parser to accept an optional positional question so `leann ask my-docs "..."` works
- reuse the provided question in interactive/non-interactive flows and add a parser regression test

## Testing
- `uv run python -m pytest tests/test_cli_ask.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,6 +546,9 @@ leann search my-docs "machine learning concepts"
 # Interactive chat with your documents
 leann ask my-docs --interactive
 
+# Ask a single question (non-interactive)
+leann ask my-docs "Where are prompts configured?"
+
 # List all your indexes
 leann list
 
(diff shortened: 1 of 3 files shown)


yichuan-w:
LGTM

PR #115, opened 2025-09-23
feat: Add support for configurable local LLM endpoints

andylizf:
## Summary
Adds support for configurable local LLM service endpoints, allowing LEANN to work with various local services beyond Ollama's default endpoint.

Fixes #114

## Changes
- Add centralized configuration module (`settings.py`) for resolving endpoints from environment variables
- Support custom hosts for Ollama-compatible services via `--host` CLI parameter  
- Support OpenAI-compatible local services via `--api-base` and `--api-key` parameters
- Pass embedding provider options through the entire pipeline (CLI → Builder → Server)
- Update both LLM chat and embedding services to use configurable endpoints

## Environment Variables
The following environment variables are now supported:
- `LEANN_OLLAMA_HOST`, `OLLAMA_HOST`, `LOCAL_LLM_ENDPOINT` - for Ollama-compatible services
- `LEANN_OPENAI_BASE_URL`, `OPENAI_BASE_URL` - for OpenAI-compatible services
- `OPENAI_API_KEY` - for API key authentication

## Example Usage
```bash
# Use MSTY Studio endpoint
leann ask my-index "What is LEANN?" --llm ollama --host http://localhost:10000

# Use LM Studio's OpenAI-compatible endpoint  
leann ask my-index "What is LEANN?" --llm openai --api-base http://localhost:1234/v1

# Build index with custom embedding endpoint
leann build my-index --docs ./docs --embedding-mode ollama --embedding-host http://localhost:10000
```

## Testing
Tests pass locally with `uv run --with '.[dev]' pytest`

🤖 Generated with [Claude Code](https://claude.ai/code)

Issue #80, opened 2025-08-31
[Feature Request]: Support LM Studio


Drjacky:
https://github.com/yichuan-w/LEANN/discussions/78#discussioncomment-14261319

MihirKohli:
Any specific details you want to mention @yichuan-w for this ticket thank you for any detail that will be provided

yichuan-w:
Sure, basically this one is into two feature requests,

1. The first one is to use LM Studio to generate embedding and LLM generation. Right now, we can only support ollama.
https://github.com/yichuan-w/LEANN/blob/main/packages/leann-core/src/leann/embedding_compute.py#L509
Here is for embedding generation, and here is for LLM generation
https://github.com/yichuan-w/LEANN/blob/main/packages/leann-core/src/leann/chat.py#L525

2. The second one is the request in 
https://github.com/yichuan-w/LEANN/discussions/78#discussion-8816528

People are concerned about their privacy when using claude code on their server, but their server do not have any GPU, so we can serve the embedding model on our local machine(let's say Macbook with GPU) and all the embedding computation operations can be posted to our local machine, so we just need to serve an endpoint(no matter ollama or LM studio) locally and when user run LEANN on server they can get the embedding results form their local machine.
Let me know if there are further questions
@MihirKohli

MihirKohli:
Thank you Taking up this feature request, will ask doubt if i will have any.

t41372:
LM Studio’s LLM and embedding API are OpenAI-compatible (and so is Ollama if you use its OpenAI-compatible endpoint). 

Since LEANN supports OpenAI, it automatically supports LM Studio and a bunch of other providers without any code changes. This definitely needs better documentation, though.

---

**How do I use LM Studio in LEANN as an LLM provider?**

```bash
export OPENAI_API_KEY="xxx"
export OPENAI_BASE_URL="http://localhost:1234/v1" # base url for lm studio by default
```

```bash
leann ask my-docs --interactive --llm openai --model qwen3-8b
```

**Same for embedding models**

```bash
export OPENAI_API_KEY="xxx"
export OPENAI_BASE_URL="http://localhost:1234/v1" # base url for lm studio by default

leann build my-docs --docs ./Example \
  --embedding-mode openai \
  --embedding-model text-embedding-qwen3-embedding-0.6b
```

As long as the model name matches what your backend exposes, you’re good to go.

---

[...]

PR #118, opened 2025-09-24
docs: update README for openai compatible llm and embedding endpoints

### LINKED ISSUE #102: Remote Embedding Server Endpoint
Some Docker containers may have no GPU drivers, but having internet access to the host machine. We want to have a configuration to avoid launching a new embedding server, instead allow to configure a valid address where a compatible embedding server listening.

t41372:
## What does this PR do?

Added instruction to let LEANN connect to any OpenAI compatible LLM and embedding backend.

Context: https://github.com/yichuan-w/LEANN/issues/80#issuecomment-3317858264

There are some other things can be done regarding documentation on LLM and embedding configuration. They are not included as a part of this pr.
- reference relevant docs in README
- add/edit docs to match new changes from #115 to README
- make sure that the rest of the documentation are consistent with the fact that LEANN actually support all open ai compatible endpoint.

## Related Issues

This should resolve #102 as we can connect to any open ai compatible inference engine remotely with instruction mentioned in the doc.

## Checklist

- [~] Tests pass (`uv run pytest`)
    - Some test failed but probably due to misconfiguration of environment on my local machine. This pr only edits docs so it shouldn't be a concern.
- [x] Code formatted (`ruff format` and `ruff check`)
- [x] Pre-commit hooks pass (`pre-commit run --all-files`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,7 +182,10 @@ LEANN supports RAG on various data sources including documents (`.pdf`, `.txt`, 
 
 ### Generation Model Setup
 
-LEANN supports multiple LLM providers for text generation (OpenAI API, HuggingFace, Ollama).
+#### LLM Backend
+
+LEANN supports many LLM providers for text generation (HuggingFace, Ollama, and Any OpenAI compatible API).
+
 
 <details>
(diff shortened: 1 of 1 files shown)


yichuan-w:
LGTM

PR #138, opened 2025-10-07
security: Enhance Hugging Face model loading security - resolves #136

### LINKED ISSUE #136: Enhance security when loading Hugging Face models

Currently, LEANN loads models from the Hugging Face Hub using the `trust_remote_code=True` parameter. This creates a potential security risk for all users, as it allows code from the model's repository to be executed on the user's machine.

Even when using LEANN for local tasks, a user might load a new model from the Hub. If that model's repository were ever compromised, it could lead to arbitrary code execution on the user's machine. This change would proactively protect users from this potential supply chain vulnerability.

### Proposed solution

I propose changing the default behavior to prioritize security by setting `trust_remote_code=False` within packages/leann-core/src/leann/chat.py.

This would prevent remote code from being executed automatically. If a specific model requires this setting to be enabled, it could be implemented as an explicit opt-in choice for the user, ideally with a warning about the security implications.

### Example usage

```python
The code in packages/leann-core/src/leann/chat.py could be updated from:

model = AutoModel.from_pretrained(model_path, trust_remote_code=True)

To:

model = AutoModel.from_pretrained(model_path, trust_remote_code=False)
```

ASuresh0524:
- Set trust_remote_code=False by default in HFChat class
- Add explicit trust_remote_code parameter to HFChat.__init__()
- Add security warning when trust_remote_code=True is used
- Update get_llm() function to support trust_remote_code parameter
- Update benchmark utilities (load_hf_model, load_vllm_model, load_qwen_vl_model)
- Add comprehensive documentation for security implications

Security Benefits:
- Prevents arbitrary code execution from compromised model repositories
- Requires explicit opt-in for models that need remote code execution
- Shows clear warnings when security is reduced
- Follows security-by-default principle

Migration Guide:
- Most users: No changes needed (more secure by default)
- Users with models requiring remote code: Add trust_remote_code=True explicitly
- Config users: Add 'trust_remote_code': true to LLM config if needed

Fixes #136 

## Checklist

### DIFF
--- a/benchmarks/llm_utils.py
+++ b/benchmarks/llm_utils.py
@@ -54,29 +54,51 @@ def extract_thinking_answer(response):
     return response.strip()
 
 
-def load_hf_model(model_name="Qwen/Qwen3-8B"):
-    """Load HuggingFace model"""
+def load_hf_model(model_name="Qwen/Qwen3-8B", trust_remote_code=False):
+    """Load HuggingFace model
+
+    Args:
+        model_name (str): Name of the model to load
+        trust_remote_code (bool): Whether to allow execution of code from the model repository.
+            Defaults to False for security. Only enable for trusted models.
+    """
     if not HF_AVAILABLE:
         raise ImportError("transformers not available")
 
+    if trust_remote_code:
+        print(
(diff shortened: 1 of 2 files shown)


yichuan-w:
LGTM

PR #147, opened 2025-10-15
Add messages regarding the use of token during query

CelineNi2:
## What does this PR do?

Adds print statements to show token usage and a warning message when the token limit is exceeded (as the API fails silently in that case).


yichuan-w:
LGTM

PR #150, opened 2025-10-23
Feature/add metadata output

ww2283:
## What does this PR do?

  Add `--show-metadata` flag to display file metadata in search results and fix ZMQ
  linking issues in Python extension build system.

  ## Related Issues

  Fixes #144

  ## Checklist

  - [x] Tests pass (`uv run pytest`)
  - [x] Code formatted (`ruff format` and `ruff check`)
  - [x] Pre-commit hooks pass (`pre-commit run --all-files`)

PR #152, opened 2025-10-23
Feature/optimize ollama batching

### LINKED ISSUE #151: [BUG] Severe Performance Bottleneck in Ollama Embedding Generation due to Serial API Calls
### What happened?

When using LEANN to compute embeddings with the Ollama provider (--embedding-mode ollama), there is a significant performance issue during the document indexing phase. The process becomes extremely slow, even for a relatively small number of documents.

Analysis of the underlying code shows that the compute_embeddings_ollama function processes text chunks serially instead of using batching, which the Ollama /api/embeddings endpoint supports and is optimized for.

Specifically, the code iterates through each text chunk and sends an individual HTTP request to the Ollama API for each one. This results in massive overhead from repeated network calls and prevents the GPU from being used efficiently, as it cannot process chunks in parallel.

### How to reproduce

Prepare a small to medium-sized codebase or a set of documents for indexing (e.g., 100-200 documents, resulting in ~2000-3000 text chunks).

Run the LEANN build/indexing command using an Ollama embedding model.

bash
leann build --embedding-model qwen3-embedding:0.6b --embedding-mode ollama
Observe the progress bar during the "Computing Ollama embeddings" step.

Expected Behavior:

The embedding process should be fast, leveraging the batch processing capabilities of the Ollama API. For a few thousand chunks, this process should complete in a few minutes at most on modern hardware, with each batch taking only a few seconds.

Actual Behavior:

The embedding process is extremely slow, taking upwards of 40-50 seconds per "batch" as shown by the progress bar. The total time is estimated to be close to an hour for just a few thousand chunks. This indicates that while the progress bar is tracking batches, the underlying API calls are being made serially for each chunk within that batch.

  Looking at compute_embeddings_ollama (lines 722-769):
```
  def get_batch_embeddings(batch_texts):
      """Get embeddings for a batch of texts."""
      all_embeddings = []
      failed_indices = []

      for i, text in enumerate(batch_texts):  # ← Sequential processing!
          # ... retry logic ...
          response = requests.post(
              f"{resolved_host}/api/embeddings",
              json={"model": model_name, "prompt": truncated_text},  # ← One text at a time
              timeout=30,
          )
```
  The problem: Even though texts are divided into "batches" (lines 784-801), each text makes a separate sequential HTTP 
  request to Ollama. This is extremely slow.

Suggested Solution:

The compute_embeddings_ollama function should be refactored to send a list of text chunks (a batch) in a single API call to the /api/embeddings endpoint. The JSON payload should contain "prompt": ["text1", "text2", "text3", ...].

This change would dramatically reduce the number of HTTP requests and allow the backend (Ollama and the GPU) to process the embeddings in parallel, leading to a significant performance improvement that aligns with user expectations for local embedding generation.

### Error message

```shell

```

### LEANN Version

0.3.4

### Operating System

macOS

ww2283:
## Related Issues

Fixes #

## Checklist

- [X] Tests pass (`uv run pytest`): as mentioned in #150 , ast chunking part is waiting to be synced with the fix in 150; diskann is not configure on my mac but this 151 is irrelevant regarding to the diskann backend. Other tests all pass
- [X] Code formatted (`ruff format` and `ruff check`)
- [X] Pre-commit hooks pass (`pre-commit run --all-files`)

 ## Summary

  Fixes #151

  This PR implements true batch processing for Ollama embeddings by migrating from the deprecated `/api/embeddings`
  endpoint to the modern `/api/embed` endpoint.

  > **Note**: This PR is based on `feature/add-metadata-output` and includes changes from PR #150 (metadata output + ZMQ
  fixes). The Ollama batching optimization is in the latest commit (8bb7743). PR #150 can be merged first, then this PR
  will cleanly apply the batching changes only.

  ## Commits in this PR

  1. `d6a3c28` - feat: add metadata output to search results (from #150)
  2. `76e1633` - fix: resolve ZMQ linking issues (from #150)
  3. `5073f31` - style: apply ruff formatting (from #150)
  4. `8bb7743` - **feat: implement true batch processing for Ollama embeddings** ⭐ (NEW)

  ## Changes (Ollama Batching)

  - **Endpoint migration**: `/api/embeddings` → `/api/embed`
  - **Parameter update**: `"prompt"` (single text) → `"input"` (array of texts)
  - **Response parsing**: Updated to handle batch embeddings array
  - **Timeout**: Increased to 60s for batch processing
  - **Error handling**: Improved for batch request failures

  ## Performance Impact

  ### Before
  - Made 32 separate API calls per "batch" of 32 texts
  - Significant HTTP connection/handshake overhead
  - Used deprecated API endpoint

  ### After
  - Makes 1 API call per batch of 32 texts
  - Reduces HTTP overhead by 32×
  - Uses modern, recommended API endpoint

  ### Actual Results (tested on 2,374 chunks)
  - API calls reduced from ~76K to ~75 requests
  - Batch size: 32 for MPS/CPU, 128 for CUDA
  - Processing time: Moderate improvement from reduced HTTP overhead

  ## Important Note

  Through empirical testing on Ollama v0.12.6, we confirmed that Ollama processes embedding batch items **sequentially**
  (not in parallel). The performance improvement comes from reduced HTTP overhead, not GPU parallelization.

  For users requiring maximum performance, we recommend e.g. :
  ```bash
  --embedding-mode sentence-transformers --embedding-model Alibaba-NLP/gte-Qwen2-1.5B-instruct

  This provides true GPU batch parallelization and is 10-20× faster than Ollama for large embedding workloads.

  Testing

  Tested with:
  - 2,374 chunks, avg 4,610 chars each
  - Ollama v0.12.6
  - Model: qwen3-embedding:0.6b
  - Batch size: 32 texts per batch

  Related Files

  - packages/leann-core/src/leann/embedding_compute.py:570-861

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,8 +3,7 @@
 	url = https://github.com/yichuan-w/DiskANN.git
 [submodule "packages/leann-backend-hnsw/third_party/faiss"]
 	path = packages/leann-backend-hnsw/third_party/faiss
-	url = https://github.com/ww2283/faiss.git
-	branch = fix/zmq-linking
+	url = https://github.com/yichuan-w/faiss.git
 [submodule "packages/leann-backend-hnsw/third_party/msgpack-c"]
 	path = packages/leann-backend-hnsw/third_party/msgpack-c
 	url = https://github.com/msgpack/msgpack-c.git


ww2283:
fix for #151

ww2283:
@ASuresh0524 

## Submodule Conflict Explanation

### Why `.gitmodules` Has a Conflict

This PR temporarily changes the faiss submodule URL from `yichuan-w/faiss` to `ww2283/faiss` because:

1. **ZMQ Linking Fix Required:** The changes in this PR depend on ZMQ linking fixes in the faiss submodule
2. **CI Was Failing:** GitHub Actions couldn't checkout the faiss commit with the fix because it wasn't pushed upstream yet
3. **Temporary Fork:** Using `ww2283/faiss:fix/zmq-linking` branch allows CI to pass while waiting for upstream merge

### Upstream PR Created

Created PR to merge ZMQ fix upstream: https://github.com/yichuan-w/faiss/pull/3

### Resolution Plan

Once the faiss PR is merged:
1. Update `.gitmodules` back to `url = https://github.com/yichuan-w/faiss.git`
2. Remove the `branch = fix/zmq-linking` line
3. Force push updated branches
4. Conflict will be resolved

### Alternative: Accept This Temporarily

If you'd prefer not to wait, this PR can be merged as-is:
- ✅ All functionality works correctly
- ✅ CI passes with the fork
- ✅ Can switch back to upstream faiss in a follow-up PR after the fix is merged

yichuan-w:
Hi @ww2283 can you change to the new Faiss version(i.e link to our backbone instead of yours) and then we can merge? I have merge the PR in faiss, thanks for the contribution!

packages/leann-core/src/leann/embedding_compute.py
```diff
@@ -717,56 +718,55 @@ def compute_embeddings_ollama(
...
-                        timeout=30,
-                    )
-                    response.raise_for_status()
+        while retry_count < max_retries:
+            try:
+                # Use /api/embed endpoint with "input" parameter for batch processing
+                response = requests.post(
+                    f"{resolved_host}/api/embed",
+                    json={"model": model_name, "input": truncated_texts},
+                    timeout=60,  # Increased timeout for batch processing
```
yichuan-w, line 737: I am wondering if this will result in OOM? 
If you test on a large scale, I think I am fine with this

yichuan-w: oh see, it is within 32 batches, then I think you are one hundred percent correct

yichuan-w:
Let's merge this PR. It is an important issue, let's merge ASAP
Thanks for your contribution!!!!

packages/leann-core/src/leann/embedding_compute.py
ww2283: I will keep this in mind when doing next step as I closely monitor its behavior. thanks for the merge! I will check around to be sure about the conflict are resolved before next step. currently Ollama has its limitation, that is the batching is correctly received but not really properly batched in itself, which is not the same behavior as in other client e.g. lm studio. So lm studio is using openai mode endpoint, and it's not oom, so I assume that ollama should be fine, even when later they decide to do proper batching. but for now the batching is ready with ollama. sadly headless server autoloading and unloading model with proper JIT is still the most smooth with ollama. Or next close solution is llama-swap but not as convenient. currently, the most speedy solution in apple silicon is either ollama with moe embedding model, which we currently only have that nomad v2, or lm studio with embeddinggemma which can offer equivalent speed comparing to that ollama hosted moe. embeddinggemma has the great two advantages: longer sequence length support (2048 vs 512) and template prepending which should theoretically be important for better results.

ww2283: on a side note, speed is important, at least to me, because I use a posttoolhook in claude code that will embed once see a git commit to keep the codebase indexing up to date. so the embedding in LEAAN has to be fast.

PR #178, opened 2025-11-26
Fix/pdf duplicate processing 175

### LINKED ISSUE #175: it run with pdf
1

ASuresh0524:
## Fix: Prevent duplicate PDF processing when using --file-types .pdf

Fixes #175

### Problem
When `--file-types .pdf` is specified, PDFs were being processed twice:
1. Separately with PyMuPDF/pdfplumber extractors (lines 1118-1163)
2. Again in the 'other file types' section via SimpleDirectoryReader (line 1184)

This caused duplicate processing and potential conflicts.

### Solution
- Exclude `.pdf` from `other_file_extensions` when PDFs are already processed separately
- Only load other file types if there are extensions to process
- Prevents duplicate PDF processing

### Changes
- Added logic to filter out `.pdf` from `code_extensions` when loading other file types if PDFs were processed separately
- Updated `SimpleDirectoryReader` to use filtered extensions
- Added check to skip loading if no other extensions to process

### Testing
Tested with `--file-types .pdf` to ensure PDFs are processed only once.

### DIFF
--- a/packages/leann-core/src/leann/cli.py
+++ b/packages/leann-core/src/leann/cli.py
@@ -1162,6 +1162,11 @@ Examples:
                             print(f"Warning: Could not process {file_path}: {e}")
 
             # Load other file types with default reader
+            # Exclude PDFs from code_extensions if they were already processed separately
+            other_file_extensions = code_extensions
+            if should_process_pdfs and ".pdf" in code_extensions:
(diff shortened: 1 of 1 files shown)


yichuan-w:
again please make sure this PR is separate for the issue and do not submit colqwen one, you should use a new branch from main
And do not change faiss

yichuan-w:
Thanks!!

PR #196, opened 2025-12-25
fix: limit depth of index discovery to prevent slow scans

### LINKED ISSUE #122: leann list scans all the things
### What happened?

I ran `leann list` in my home directory.  An hour later, I killed it.  It looks like it's scanning everything under $HOME.  Seems to be a bit much.  Makes the tool unusable from a point in the filesystem where there's a lot of files.

### How to reproduce

Run `leann list` in $HOME.

### Error message

```shell

```

### LEANN Version

0.1.0

### Operating System

macOS

majiayu000:
## Summary

- Fixes the issue where `leann list` scans all files in large directories like $HOME
- Adds `_find_meta_files_limited()` method with configurable max_depth (default: 3)
- Skips common large directories (node_modules, .venv, .git, __pycache__, etc.)
- Applies limited search in both CLI and registry modules

Fixes #122

## Test plan

- [x] Added 8 new tests in `tests/test_cli_list_performance.py`
- [x] Tests verify depth limiting works correctly
- [x] Tests verify skip directories are properly excluded
- [x] All new tests pass locally

### DIFF
--- a/packages/leann-core/src/leann/cli.py
+++ b/packages/leann-core/src/leann/cli.py
@@ -404,6 +404,68 @@ Examples:
             absolute_path = Path(str(file_path))
         return gitignore_matches(absolute_path.as_posix())
 
+    def _find_meta_files_limited(
+        self, root: Path, max_depth: int = 3, pattern: str = "*.leann.meta.json"
+    ):
+        """Find meta files with limited depth to avoid scanning large directories.
+
+        Skips common large directories that shouldn't contain LEANN indexes.
+        """
+        # Directories to skip - these are typically large and won't contain user indexes
+        skip_dirs = {
+            ".git",
+            ".svn",
+            ".hg",
+            "node_modules",
+            "__pycache__",
+            ".venv",
+            "venv",
+            ".env",
+            "env",
+            ".tox",
+            ".nox",
+            ".mypy_cache",
+            ".pytest_cache",
+            ".ruff_cache",
+            "dist",
+            "build",
+            ".eggs",
+            "*.egg-info",
+            ".cache",
+            ".npm",
+            ".yarn",
+            "vendor",
+            "Pods",
+            ".gradle",
+            "target",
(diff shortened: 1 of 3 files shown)


andylizf:
Thanks for working on this! A few suggestions:

## 1. Add `--max-depth` CLI option

Instead of hardcoding `max_depth=3`, consider adding a CLI option with a reasonable default:

```bash
leann list --max-depth 5
```

3 levels might be too shallow for some project structures (e.g., `project/data/experiments/v1/index.leann` would be missed).

## 2. Consider a design improvement: Full registry instead of scanning

Currently the code scans for "Apps format" indexes (`*.leann.meta.json`) because they can be anywhere in the project. But this is the root cause of the performance issue.

A better approach might be to **register all indexes** (both CLI and App-created) in the global registry with their full paths:

```json
// ~/.leann/indexes.json
{
  "my-docs": "/path/to/project/.leann/indexes/my-docs",
  "app-index": "/path/to/project/data/app-index.leann"
}
```

This way `leann list` becomes O(1) - just read the registry, no scanning needed at all.

This could be a follow-up issue/PR, but worth considering as the long-term solution.

majiayu000:
Thanks for the feedback @andylizf!

I've added the `--max-depth` CLI option as suggested. Users can now customize the scan depth:

```bash
# Default depth (3)
leann list

# Scan deeper directories
leann list --max-depth 5
```

For the global registry suggestion (making `leann list` O(1)), I agree it's a great long-term improvement. I'll create a follow-up issue to track that enhancement.

majiayu000:
**Update:** I've also implemented the long-term global registry solution in a separate PR #199.

- **This PR (#196)** - Short-term fix: adds `--max-depth` CLI option
- **PR #199** - Long-term solution: implements O(1) registry lookup (closes #198)

This PR can be merged independently. PR #199 builds on top of this one.

PR #214, opened 2026-01-12
Fix Gemini OpenAI-compat embedding batch size limit

hummat:
Gemini's OpenAI-compatible embedding endpoint rejects >100 inputs per request ("BatchEmbedContentsRequest.requests: at most 100 requests can be in one batch").

This PR:
- Caps OpenAI-compat embedding batch size to 100 when base_url points at generativelanguage.googleapis.com.
- Removes stray debug prints.
- Adds a unit test verifying chunking [100, 100, 50] for 250 inputs.

Repro:
```bash
leann build my-index ./docs \
  --embedding-mode openai \
  --embedding-api-base https://generativelanguage.googleapis.com/v1beta/openai \
  --embedding-model gemini-embedding-001
```

### DIFF
--- a/packages/leann-core/src/leann/embedding_compute.py
+++ b/packages/leann-core/src/leann/embedding_compute.py
@@ -778,7 +778,10 @@ def compute_embeddings_openai(
     #   "BatchEmbedContentsRequest.requests: at most 100 requests can be in one batch"
     if "generativelanguage.googleapis.com" in (resolved_base_url or ""):
         max_batch_size = min(max_batch_size, 100)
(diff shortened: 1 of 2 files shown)


ASuresh0524:
@hummat Hey! Would it be possible to fix the CI error feel as this PR may be helpful for the general community? Thanks for the help :)

hummat:
@ASuresh0524 can you take a look? I think this one isn't on me :)

ASuresh0524:
Yeah, don't think it is lol, going to ping some others for help, don't know how to fix on my end. @yichuan-w @andylizf any thoughts?

Issue #60, opened 2025-08-16
[Feature]Code retrieval enhancements Merkle Tree 2/N


yichuan-w:
Let's Learn from this https://github.com/search?q=repo%3Azilliztech%2Fclaude-context+merkle&type=code
Also, if you are interested in this, let us schedule a meeting and talk to the Hugging Face people

yichuan-w:
Hi, for anyone who is interested, the following steps is:

1. Add a Merkle tree to track file-level change, if they are change, we can rechunk and re-embedd the whole file
2. We should dis-able recompute in this setting. since recompute is hard here
3. The API should like 
LEANN watch following mixbread to check if there are file changes

aloumakos:
I'd like to work on this, if it's still relevant.

yichuan-w:
Sure,  it is still very important!

yichuan-w:
And built upon this, we can add operations like leann watch to trace whether the file has been changed, if this file has been changed, delete old embeddings and add new embeddings,
(We still need to check the add/delete operation if implemented correctly)
If I understand correctly, add w recompute is added but not that largely tested, add wo recompute should be good to use out of box.
delete is somehow tricky, I guess, for HNSW even Faiss does not have delete.
Some options: 1. soft delete, ignore the results of delete node and search more
2. Use IVF
or check Faiss again.

Issue #61, opened 2025-08-16
[Feature]Add pgvector support for frequent update 3/N


yichuan-w:
Right now, we only support leann backend(which can fall back to faiss easily), but if we want to frequently update, we need pgvector or Qdrant for the frequent update case

MihirKohli:
Looking forward to work on this feature can you please add more details

Thank you

yichuan-w:
Hi @MihirKohli , this one is a little but tricy right now and I am working on that, so can you pick some other PR to work around like #80,#36, #47, 47 should be intersting

MihirKohli:
sure let me take a look

yichuan-w:
Some follow up [discussion](https://github.com/yichuan-w/LEANN/pull/219#issuecomment-3766151132).

yichuan-w:
#242  follow up PR here

PR #219, opened 2026-01-17
feat: add file syncing logic using a merkle tree

aloumakos:
## What does this PR do?

- Introduce a flat merkle tree class to easily check for file changes
- Exposes an API for file syncing that checks for new changes and handles snapshot storage
- Unit tests for the above

## Related Issues

Fixes [#60](https://github.com/yichuan-w/LEANN/issues/60)

## Checklist

- [x] Tests pass (`uv run pytest`)
- [x] Code formatted (`ruff format` and `ruff check`)
- [x] Pre-commit hooks pass (`pre-commit run --all-files`)

### DIFF
--- a/tests/test_sync.py
+++ b/tests/test_sync.py
@@ -49,8 +49,8 @@ class TestMerkleTreeCompare(unittest.TestCase):
 
         added, removed, modified = MerkleTree.compare_with(tree1, tree2)
 
-        self.assertEqual(added, ["b.txt"])
-        self.assertEqual(removed, ["c.txt"])
+        self.assertEqual(added, ["c.txt"])
+        self.assertEqual(removed, ["b.txt"])
         self.assertEqual(modified, ["a.txt"])
 
 


yichuan-w:
That is a great PR I will keep following this.
And built upon this, we can add operations like `leann watch` to trace whether the file has been changed, if this file has been changed, delete old embeddings and add new embeddings,
(We still need to check the add/delete operation if implemented correctly)
If I understand correctly, add w recompute is added but not that largely tested, add wo recompute should be good to use out of box.
delete is somehow tricky, I guess, for HNSW even Faiss does not have delete.
Some options: 1. soft delete, ignore the results of delete node and search more
2. Use IVF
or check Faiss again.

yichuan-w:
cc @aloumakos , do you have time follow up on this?

aloumakos:
As you said, faiss's HNSW does not provide API for deletions, so we would need to look elsewhere. I thought that [#61](https://github.com/yichuan-w/LEANN/issues/61) would provide a suitable backend, if that (or IVF) is implemented then a leann.watch API should not be too hard to write.

By the way, it seems that https://github.com/nmslib/hnswlib supports flagging records as deleted and is able to replace them with newly added items

yichuan-w:
Yeah, I guess now IVF is not that well-supported, we need either pgvectors/hnswlib/IVF in faiss to support that , or we can self write sof-deletion but take some effort, we can follow up on this feature!
And I will prioritize this one compared to other PR and if you want , we can discuss more or you can submit which you love.

yichuan-w:
https://github.com/yichuan-w/LEANN/pull/242
@aloumakos 
I use your interface to implement 'leann watch', hopefully we can monitor our code by leann in near future

PR #242, opened 2026-02-11
feat: add `leann watch` command, line numbers in code search, and fix search output

yichuan-w:
## Summary

- **`leann watch` command**: Compares the current file system state against the last Merkle tree checkpoint and reports added/removed/modified files along with their associated chunk IDs. Integrates with `leann build` to automatically create initial snapshots and persist sync config.
- **Line numbers in code chunks**: Code files now have line numbers prepended to each line (e.g. `42|def foo():`) during chunking, so search results show exact line locations for easy code navigation. Partial first lines from chunk overlap are trimmed to keep output clean.
- **Fix search output being swallowed**: `suppress_cpp_output` was redirecting OS-level stdout/stderr to `/dev/null` to silence HNSW/FAISS C++ logs, but this also silenced Python `print()`. Fixed by pointing `sys.stdout`/`sys.stderr` at saved fd copies so Python output reaches the terminal while C++ output is still suppressed.
- **README updates**: Added `leann watch` documentation to the feature list, Usage Examples, and Complete CLI Reference.

## Test plan

- [ ] `leann build test-idx --docs ./some-dir --no-recompute` creates index + `sync_roots.json` + `.sync_context.pickle`
- [ ] `leann watch test-idx` shows "No changes detected" immediately after build
- [ ] Modify a file, then `leann watch test-idx` shows it as modified with chunk IDs
- [ ] `leann search test-idx "query" --no-recompute` displays search results (not swallowed) with line numbers in code chunks
- [ ] Verify code chunks start at clean line boundaries (no partial first lines)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,3 +109,5 @@ apps/multimodal/vision-based-pdf-multi-vector/multi-vector-colpali-native-weavia
 
 # AUR build directory (Arch Linux)
 paru-bin/
+merkle-tree-test/
+*.pickle
--- a/README.md
+++ b/README.md
@@ -1082,6 +1082,9 @@ leann ask my-docs --interactive
 # Ask a single question (non-interactive)
 leann ask my-docs "Where are prompts configured?"
 
+# Detect file changes since last build/watch checkpoint
+leann watch my-docs
+
 # List all your indexes
 leann list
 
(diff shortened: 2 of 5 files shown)


andylizf:
@yichuan-w Seems this PR introduces some regressions. Can you help investigate it?

yichuan-w:
fixed in #245 @andylizf

PR #246, opened 2026-02-12
Gate releases on successful main CI

andylizf:
## Summary
- block manual release when latest main commit CI is not successful
- add a preflight check in the release workflow


andylizf:
@yichuan-w The merged commit that failed CI was 49bc4d1435e5c177c88c98ff28a0c2fb25c39b0f (from PR #242). This guard would have blocked that release path.

Issue #182, opened 2025-12-05
LEANN search latency problem


yichuan-w:
This is the log of running claude code, it can run stably, but it do have some problems of the embedding server start problem, which we might need to solve.  cc @andylizf @ASuresh0524 
https://github.com/yichuan-w/LEANN/blob/main/packages/leann-mcp/README.md

Here is the log
```
(vllm) ➜  vllm git:(main) ✗ claude               

> /mcp 
  ⎿  MCP dialog dismissed

> please tell me whats the page size in vllm use leann search to localize 


leann-server - leann_search (MCP)(index_name: "my-repo", query: "page size configuration block size", top_k: 10, show_metadata: true, complexity: 32)
  ⎿  ⚠️ Embedding server failed: Failed to compute embeddings via server: Resource temporarily unavailable                                                                                                                   
     ⏭️ Falling back to direct model loading...
     Search results for 'page size configuration block size' (top 10):
     … +139 lines (ctrl+o to expand)

● Great! I found several relevant files. Let me read the key files to understand the page size configuration:

● Read(vllm/v1/core/kv_cache_utils.py)
  ⎿  Read 1473 lines

● Search(pattern: "block_size|page_size|BLOCK_SIZE", path: "vllm", type: "py", output_mode: "content", head_limit: 30)
  ⎿  Found 30 lines (ctrl+o to expand)

● Let me search for more specific information about the default page size value:

● leann-server - leann_search (MCP)(index_name: "my-repo", query: "default page size block size 16 configuration cache config", top_k: 10, show_metadata: true, complexity: 32)
  ⎿  ⚠️ Embedding server failed: Failed to compute embeddings via server: Resource temporarily unavailable                                                                                                                   
     ⏭️ Falling back to direct model loading...
     Search results for 'default page size block size 16 configuration cache config' (top 10):
     … +138 lines (ctrl+o to expand)

```

yichuan-w:
Related to #177

PR #248, opened 2026-02-13
Fix cold-start latency + ZMQ reliability

### LINKED ISSUE #177: Search with `recompute` second level latency for code RAG

raoabinav:
﻿Closes #177, #182

**What changed and why:**
- **Removed redundant `_ensure_server_running` call inside `compute_query_embedding`.** Previously, every query embedding triggered a second server health check even though the caller (`api.py`) already called `_ensure_server_running`. This was the main source of cold-start latency â€” a double server startup on first query. Now `compute_query_embedding` trusts the port it receives from the caller, eliminating the redundant check.
- **Added ZMQ retry with exponential backoff (0.5s, 1s, 2s) in `_compute_embedding_via_server`.** The original code did a single ZMQ send/recv with no retry â€” any transient connection failure (server still loading, socket not yet bound) was a hard crash. Now retries up to 3 times with proper socket teardown between attempts.
- **Set `ZMQ.SNDTIMEO=10s` and `ZMQ.LINGER=0`.** Without SNDTIMEO, a dead server caused an indefinite hang on `socket.send()`. Without LINGER=0, `socket.close()` would block waiting for unsent messages, stalling the retry loop.
- **`enable_warmup` is now a kwargs pop, not a forward.** It was being passed through to `start_server()` which didn't understand it. Now it's consumed by `_ensure_server_running` to fire a dummy embedding request after server start, pre-loading the model into GPU memory before the first real query hits.
- **Replaced `print()` with `logging.getLogger(__name__)` in searcher_base.** The original used bare `print("âš ï¸ ...")` for error reporting, which is invisible in production and breaks structured logging pipelines.
- **Added timing instrumentation across the entire server lifecycle** â€” `start_server`, `_start_new_server`, `_ensure_server_running`, `compute_query_embedding` all log elapsed time so you can actually profile where cold-start time goes.

### DIFF
--- a/scripts/fix-ci-all-branches.ps1
+++ b/scripts/fix-ci-all-branches.ps1
@@ -1,0 +1,47 @@
+# Fix CI (lint/ruff) on all 8 PR branches
+# Run from repo root - matches CI: pre-commit run --all-files
+
+$ErrorActionPreference = "Continue"
+$branches = @(
+    "abinav/issue-177-cold-start-LONFb",
+    "abinav/issue-233-hybrid-search-LONFb",
+    "abinav/issue-141-reindex-LONFb",
+    "abinav/issue-158-ocr-LONFb",
+    "abinav/issue-217-llamaindex-LONFb",
+    "abinav/issue-166-warmup-LONFb",
+    "abinav/issue-96-obsidian-LONFb",
+    "abinav/issue-47-local-cursor-LONFb"
+)
+
+$repoRoot = (git rev-parse --show-toplevel 2>$null)
+if (-not $repoRoot) { Write-Error "Not in git repo"; exit 1 }
+Set-Location $repoRoot
+
+foreach ($branch in $branches) {
+    Write-Host "`n========== $branch ==========" -ForegroundColor Cyan
+    git checkout $branch 2>&1 | Out-Null
+    if ($LASTEXITCODE -ne 0) {
+        Write-Host "  Skip (branch not found)" -ForegroundColor Yellow
+        continue
+    }
+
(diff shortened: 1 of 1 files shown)


raoabinav:
Closing PR.

PR #249, opened 2026-02-13
feat: add hybrid search via BM25 (SQLite FTS5) + score fusion

### LINKED ISSUE #233: Hybrid search support[important!]

https://github.com/tobi/qmd
This repo is a good reference for hybrid search, we should try to implement BM25 in LEANN and make it to increase accuracy too!
We have https://github.com/yichuan-w/LEANN/pull/211 before, but have not provide document about how to use and how to combine together with dense search, it only provide a simple API now, also it seems that PR use a self written version of BM25(in python)

### Proposed solution

Can we use similiar method like
https://github.com/tobi/qmd?tab=readme-ov-file#search-backends
SQLite FTS5 BM25
to make it more robust

### Example usage

```python
search(query, dense-score-ratio=0.6, full-text-score-ration=0.4)
Let's make the first version of the PR and set default ratio ull text search to be 0, and when it is stable we can change that.
Note: fulltextsearch == BM25 == sparse search here
```

raoabinav:
﻿Closes #233

**What changed and why:**
- **New `bm25.py` â€” SQLite FTS5-backed BM25 index.** Chose FTS5 over a pure-Python BM25 implementation (like `rank_bm25`) because: (1) FTS5 handles tokenization/stemming natively, (2) it scales to millions of chunks without loading everything into memory, (3) it's zero-dependency (sqlite3 is in the stdlib), and (4) the BM25 ranking is built into the FTS5 `rank` function, so no custom scoring math needed.
- **`build_fts5_index()` is called at index build time** â€” added to both the `build_index()` and `build_from_embeddings()` paths in `api.py`. The FTS5 DB is stored alongside the LEANN index as `<name>.fts5.db`. For pre-computed embedding imports (where chunk text may be placeholder), it skips FTS5 construction via `skip_if_placeholder`.
- **New `hybrid.py` â€” weighted score fusion.** Implements min-max normalization on both vector distances and BM25 scores, then linearly combines them: `final = (1 - ratio) * vector_norm + ratio * bm25_norm`. Chose linear fusion over reciprocal rank fusion (RRF) because RRF ignores score magnitude and LEANN's HNSW distances are meaningful. The `sparse_score_ratio` parameter (0.0 = pure vector, 1.0 = pure BM25) is exposed through the search API.
- **BM25 index is optional** â€” if the `.fts5.db` file doesn't exist (e.g., old indexes built before this PR), search falls back to pure vector. No migration required.


raoabinav:
Closing PR.

PR #250, opened 2026-02-13
Add leann reindex CLI command with file manifest tracking

### LINKED ISSUE #141: [FEAT] Frequent update recommend no-recompute and impl leann --reindex

This is a feature widely requested

### Proposed solution

impl leann reindex

### Example usage

```python

```

raoabinav:
﻿Closes #141

**What changed and why:**
- **New `leann reindex <name> --docs <dir>` command.** The core problem: after adding/editing files in your docs directory, you had to `leann remove` + `leann build` to update the index. Now `reindex` handles it as a single operation.
- **SHA-256 file manifest for delta detection.** On every `build` or `reindex`, a `manifest.json` (mapping `{absolute_path: sha256}`) is saved alongside the index. On subsequent reindex, the manifest is diffed to report exactly which files are new/changed/deleted/unchanged. This is a content-hash approach (not mtime) so it's resilient to filesystem clock skew and `touch` operations.
- **`LeannBuilder.from_meta()` classmethod** reconstructs a builder from an existing index's `meta.json`. This avoids the user having to re-specify `--backend`, `--embedding-model`, `--graph-degree`, etc. on reindex â€” it reads the original build config and reuses it exactly.
- **Currently does a full rebuild, not incremental embedding.** Even though delta detection identifies unchanged files, the rebuild still re-embeds everything. True incremental embedding (appending to an existing HNSW graph) would require the backend to support `add_items()` which HNSW doesn't cleanly expose. The manifest delta is still valuable: it gives users visibility into what changed and skips rebuild entirely when nothing changed.
- **`--sync` flag for delete behavior.** By default, deleted files keep their chunks in the index (safe for rollbacks). `--sync` removes them. This mirrors how Elasticsearch reindex handles deletes.

### DIFF
--- a/tests/test_reindex.py
+++ b/tests/test_reindex.py
@@ -26,6 +26,7 @@ if "leann_backend_hnsw" not in sys.modules:
 def _hnsw_backend_available() -> bool:
     try:
         from leann.registry import BACKEND_REGISTRY
+
         return "hnsw" in BACKEND_REGISTRY
     except Exception:
         return False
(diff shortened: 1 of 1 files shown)


raoabinav:
Closing PR.

PR #251, opened 2026-02-13
Add --enable-ocr flag for scanned PDF support

### LINKED ISSUE #158: [feat] OCR based application

Right now, LEANN is using text embedding only. We have two other options for multimodal data:
1. Use DeepSeek OCR or [MinerU](https://github.com/opendatalab/MinerU) to process all into text space
2. maintain both image vectors and text vectors separately

### Proposed solution

RAGanything repo, MinerU

### Example usage

```python
To RAG over vision-rich task
```

raoabinav:
﻿Closes #158

**What changed and why:**
- **`--enable-ocr` flag on `leann build`.** When set, PDFs with pages containing < 50 chars of embedded text are sent through OCR. The 50-char threshold is intentional â€” it avoids OCR overhead on text-native PDFs while catching scanned documents and image-heavy slides.
- **Chose MinerU (`magic-pdf`) over pytesseract.** MinerU works at the document level (not page-by-page), understands PDF layout structure, and outputs clean Markdown with heading hierarchy preserved. Pytesseract gives flat text with no structure, requires Tesseract system binary, and is significantly slower on multi-page docs. MinerU is optional â€” `pip install leann-core[ocr]` pulls it in.
- **Fallback chain:** MinerU â†’ pymupdf built-in OCR â†’ raw text extraction. If MinerU fails or isn't installed, it falls through gracefully. The function signature of `extract_pdf_text_with_pymupdf` is backward compatible (new `use_ocr=False` kwarg).
- **`apps/ocr_rag.py` example app** showing the full pipeline: directory of scanned PDFs â†’ OCR extraction â†’ chunking â†’ index â†’ query. This extends the existing `BaseRAGExample` pattern used by other example apps.


raoabinav:
Closing PR.

PR #253, opened 2026-02-13
feat: add persistent embedding daemon to eliminate cold-start latency

### LINKED ISSUE #166: [Feature] Better warmup strategy
Fix the long warmup problem in #159

raoabinav:
﻿Closes #166

**What changed and why:**
- **New `leann serve` command** that starts a long-running background process keeping the embedding model loaded in memory. The first `leann search` after boot normally takes 10-30s (model load + tokenizer init). With the daemon pre-warmed, it's < 100ms.
- **Daemon lifecycle: `--foreground`, `--stop`, `--status`.** Writes a PID file to `~/.leann/daemon.pid` and state to `~/.leann/daemon.json`. `--stop` sends SIGTERM and cleans up the PID file. `--status` reads the state file and checks if the process is still alive.
- **`EmbeddingServerManager.connect_to_daemon()` added** â€” before starting a new server subprocess, the manager now checks if a daemon is already running on the expected port and reuses it. This is the key integration point: existing `leann search` commands transparently benefit from the daemon without any code changes in the search path.
- **Chose process-level daemon over thread-level** because the embedding model (especially sentence-transformers with PyTorch) holds a GIL-heavy computation path. A separate process avoids GIL contention with the ZMQ server's event loop. The daemon uses the same ZMQ protocol as the subprocess-based server â€” zero changes to the client side.
- **Relationship with PR 1 (cold-start fix):** PR 1 fixes the cold-start for one-off queries (retry + no double-check). This PR eliminates cold-start entirely for repeated usage (daemon stays warm). They're complementary â€” PR 1 is the safety net, this is the performance optimization.

### DIFF
--- a/packages/leann-core/src/leann/cli.py
+++ b/packages/leann-core/src/leann/cli.py
@@ -1724,10 +1724,10 @@ Examples:
             else:
                 import datetime
 
-                started = datetime.datetime.fromtimestamp(
-                    state.get("started_at", 0)
-                ).strftime("%Y-%m-%d %H:%M:%S")
-                print(f"Embedding daemon is running:")
+                started = datetime.datetime.fromtimestamp(state.get("started_at", 0)).strftime(
+                    "%Y-%m-%d %H:%M:%S"
+                )
+                print("Embedding daemon is running:")
                 print(f"  PID:    {state.get('pid')}")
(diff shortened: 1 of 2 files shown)


raoabinav:
Closing PR.

PR #255, opened 2026-02-13
feat: add local Cursor proxy - local LLM + LEANN code retrieval

### LINKED ISSUE #47: [Feature] Fully local cursor reproduce 1/N using your own ollama
https://www.reddit.com/r/LocalLLaMA/comments/1m118is/use_claudecode_with_local_models/

reference

If we set up properly, we can.  Just need some setup to serve an ollama backbone and do the index ourselves as cursor just announced their cli yesterday https://x.com/cursor_ai/status/1953559384531050724

raoabinav:
﻿Closes #47

**What changed and why:**
- **New `leann cursor` command** starts an OpenAI-compatible HTTP proxy (`/v1/chat/completions`, `/v1/models`) that intercepts chat requests, runs LEANN retrieval against a local code index, injects the retrieved context into the system prompt, and forwards to a local LLM (Ollama/LM Studio).
- **Architecture decision: HTTP proxy, not a LlamaIndex agent.** A proxy is editor-agnostic â€” Cursor, Continue, Claude Code, any OpenAI-compatible client works by just pointing `OPENAI_BASE_URL` at it. An agent-based approach would lock you into one framework.
- **Context injection strategy: prepend to system message.** If the conversation already has a system message, context is appended to it. If not, a new system message is inserted at position 0. This preserves the original system prompt's instructions while adding retrieved code. Alternative considered (injecting as a separate user message) was rejected because it pollutes the conversation history and confuses multi-turn context.
- **Uses `ThreadingHTTPServer` from stdlib** â€” no FastAPI/Flask dependency. For a local proxy that handles one user's requests, stdlib threading is sufficient and keeps the dependency footprint at zero.
- **10 MB request body limit** prevents OOM from malformed requests. CORS headers are set for browser-based editors. `_forward_to_llm` preserves upstream error bodies so debugging LLM connection issues is transparent.
- **`_build_context_block` truncates at `--max-context` chars** (default 8000) to stay within the local model's context window budget. It includes file path headers per snippet so the model can reference sources.
- **Works without an index** (`--index` is optional) â€” in that case it's a pure model proxy, useful for just routing Cursor to a local LLM without retrieval.


raoabinav:
Closing PR.

PR #271, opened 2026-02-26
refactor: idempotent build, auto-rebuild watch, and search --json

andylizf:
## Summary

Refactors `leann build` to be **idempotent**: running it multiple times always produces a correct, up-to-date index without requiring `--force`. Also upgrades `leann watch` from a one-shot change reporter to a continuous auto-rebuild loop, and adds `leann search --json` for machine-readable output.

---

### `leann build` is now idempotent (the core change)

**Before:** Users had to manually decide between `leann build` (first time) vs incremental update (new files only) vs `--force` (rebuild). If files were deleted or modified, the CLI printed an error and told you to use `--force`. Confusing.

**After:** `leann build my-docs --docs ./data/` always does the right thing:

| Situation | What happens | Speed |
|-----------|-------------|-------|
| No changes since last build | Prints "Index up to date." and exits | Instant |
| Only new files added | **Incremental update** — appends new chunks to existing index | Fast |
| Files modified or removed | **Automatic full rebuild** — no `--force` needed | Slow (unavoidable) |
| Embedding model/mode changed | Full rebuild (detects config drift) | Slow |
| Legacy index (pre-manifest) | Full rebuild + enables future incrementals | Slow (one-time) |

The CLI now prints diagnostic messages explaining which path it chose and why:

```
# Scenario: files modified, can't do incremental
Incremental update not possible (2 file(s) modified); falling back to full rebuild.
Full rebuild starting (~2 modified)...
Building index 'my-docs' with hnsw backend...

# Scenario: only new files, compact index blocks incremental
Incremental update not possible: index is compact (read-only); rebuild with
--no-compact to enable incremental updates. Falling back to full rebuild.
```

### `--compact` default changed: `True` → `False`

**This is a breaking change for users who relied on the default.**

- Compact (CSR) indices are read-only and block incremental updates
- Non-compact indices still get 97% storage compression via `--recompute` (embedding pruning)
- The only practical difference: non-compact indices support `update_index()` for appending new data

**Migration:** If you need compact indices, pass `--compact` explicitly. Existing compact indices are unaffected (they continue to work for search). Only new builds use the new default.

### `leann watch` — auto-rebuild on changes

**Before:** `leann watch my-docs` scanned files once, printed a diff-like report, and exited.

**After:** It runs as a continuous polling loop:

```bash
leann watch my-docs                  # poll every 5s, auto-rebuild on changes
leann watch my-docs --interval 10    # poll every 10s
leann watch my-docs --once           # check once, rebuild if needed, exit
leann watch my-docs --dry-run        # report changes only (old behavior)
leann watch my-docs --once --dry-run # one-shot report, no rebuild
```

The auto-rebuild reads the original build config from `meta.json` so it uses the same backend, embedding model, and options as the initial build.

### `leann search --json`

```bash
leann search my-docs "what is LEANN?" --json
```

Outputs a JSON array of `{id, score, text, metadata}` objects. Useful for MCP servers, scripts, and CI pipelines.

### Other changes

- Extracted `_build_embedding_options()` helper (eliminates duplicated embedding config code between incremental and full build paths)
- Removed emoji from search output (`📄`→`File:`, etc.) for Windows console compatibility
- Refactored `watch_index` into `_watch_check_changes` / `_watch_report_changes` / `_watch_trigger_build` for testability

## Breaking changes

| Change | Impact | Migration |
|--------|--------|-----------|
| `--compact` default `True` → `False` | New builds produce non-compact (updatable) indices | Pass `--compact` explicitly if you need read-only CSR format |
| `leann watch` now rebuilds by default | Was report-only, now triggers builds | Use `--dry-run` for old behavior |

## Test plan

- [x] All existing `test_incremental_build.py` unit tests pass
- [x] CLI argument parsing verified for all new flags
- [x] `ruff check` and `ruff format` clean
- [ ] Manual: build → add file → re-run build → incremental update
- [ ] Manual: build → modify file → re-run build → auto full rebuild
- [ ] Manual: build → delete file → re-run build → auto full rebuild
- [ ] Manual: `leann watch --once` detects changes and rebuilds
- [ ] Manual: `leann watch --dry-run` reports without rebuilding
- [ ] Manual: `leann search --json` outputs valid JSON

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -14,6 +14,14 @@ jobs:
       - uses: actions/checkout@v4
       - uses: lycheeverse/lychee-action@v2
         with:
-          args: --no-progress --insecure --user-agent 'curl/7.68.0' --exclude '.*api\.star-history\.com.*' --accept 200,201,202,203,204,205,206,207,208,226,300,301,302,303,304,305,306,307,308,503 README.md docs/ apps/ examples/ benchmarks/
+          args: >-
+            --no-progress --insecure
+            --user-agent 'curl/7.68.0'
+            --max-retries 3
+            --retry-wait-time 5
+            --exclude '.*star-history\.com.*'
+            --accept 200,201,202,203,204,205,206,207,208,226,300,301,302,303,304,305,306,307,308,503
+            README.md docs/ apps/ examples/ benchmarks/
+          fail: false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/packages/leann-core/src/leann/cli.py
+++ b/packages/leann-core/src/leann/cli.py
@@ -1892,15 +1892,27 @@ Examples:
                 elif not can_incremental:
                     blockers = []
                     if meta.get("backend_name") not in ("hnsw", "ivf"):
-                        blockers.append(f"backend '{meta.get('backend_name')}' does not support incremental updates")
-                    if meta.get("is_compact", meta.get("backend_kwargs", {}).get("is_compact", True)):
-                        blockers.append("index is compact (read-only); rebuild with --no-compact to enable incremental updates")
+                        blockers.append(
+                            f"backend '{meta.get('backend_name')}' does not support incremental updates"
+                        )
+                    if meta.get(
+                        "is_compact", meta.get("backend_kwargs", {}).get("is_compact", True)
+                    ):
+                        blockers.append(
+                            "index is compact (read-only); rebuild with --no-compact to enable incremental updates"
+                        )
                     if meta.get("embedding_model") != args.embedding_model:
-                        blockers.append(f"embedding model changed ('{meta.get('embedding_model')}' -> '{args.embedding_model}')")
+                        blockers.append(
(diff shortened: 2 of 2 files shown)


yichuan-w:
Actually, recompute and update seems in-compatible now. If we want to support update first, then we go IVF; if we support storage-efficient first, then we go LEANN. What's your idea here?

yichuan-w:
@andylizf fix the compact default setting, and leave me the following step to finish updating and making a truly incremental MCP for coding

andylizf:
Done, `--compact` defaults to `False` now.

This shouldn't be noticeable to users. The only behavioral difference is that indices become updatable by default. Storage-wise the difference is minimal either way, and anyone who prefers compact can still pass `--compact`.

This PR only touches the CLI/semantic layer (change detection, incremental vs full rebuild). Could you help double-check that `update_index` works correctly for **IVF**? File removal currently falls back to full rebuild, which still needs work on the backend side.

Happy to hand off the incremental MCP work to you!

yichuan-w:
Sure — I’ll move forward with making the MCP fully end-to-end functional (especially the remove part). For the storage-efficient HNSW backend, you can continue working on that — for example, testing add-only at large scale.

PR #303, opened 2026-04-16
fix(server): update default host to 127.0.0.1 for better security

gliptak:
## What does this PR do?

fix(server): update default host to 127.0.0.1 for better security

## Related Issues

#267 

Fixes #

## Checklist

### DIFF
--- a/packages/leann-core/src/leann/cli.py
+++ b/packages/leann-core/src/leann/cli.py
@@ -702,7 +702,7 @@ Examples:
             "serve", help="Start HTTP API server for LEANN vector DB"
         )
         serve_parser.add_argument(
-            "--host", type=str, default=None, help="Host to bind to (default: 0.0.0.0)"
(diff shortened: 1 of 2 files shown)


gliptak:
@yichuan-w please review

yichuan-w:
cc @ASuresh0524 BTW I found there are multiple places we use 127.0.1, so I am hesitant to merge this, maybe we should change all

ASuresh0524:
i agree, if you think 127.0.1 is better for infra stuff then we should change all needs of it before merging

gliptak:
what are the other code locations to update?

andylizf:
Thanks @gliptak. Good catch!

andylizf:
> cc @ASuresh0524 BTW I found there are multiple places we use 127.0.1, so I am hesitant to merge this, maybe we should change all

I think they're the only places we use `0.0.0.0`.

image

Merging this. Thanks @gliptak for the contribution. That's a great one.

PR #305, opened 2026-04-22
test: verify daemon isolation across different passages files (fixes #281)

### LINKED ISSUE #281: Daemon embedding server reuse causes DiskANN search failures across tests
## Summary

When multiple DiskANN tests run sequentially, the daemon embedding server from a previous test is incorrectly reused by the next test, even though the passages file has changed. This causes `Failed to fetch embeddings` errors in the C++ search layer.

## Symptoms

```
ERROR - Both protobuf and msgpack parsing failed!
ZMQ_FETCH_ERROR: Server response has invalid dimensions size.

RuntimeError: [FUNC: cached_beam_search][FILE: pq_flash_index.cpp][LINE: 2684]  Failed to fetch embeddings
```

Each test passes when run in isolation, but fails when run after another DiskANN test.

## Root Cause

The `EmbeddingServerManager` daemon mode (`embedding_server_manager.py`) is designed to keep embedding servers running across searches for performance. However, when a new test creates a new index with a different passages file:

1. Test A builds index A, starts daemon server on port 5558 with `passages_file_A`
2. Test A completes, but daemon server keeps running (by design, for TTL reuse)
3. Test B builds index B with different documents
4. Test B's `LeannSearcher` calls `_ensure_server_running()`, which finds the daemon on port 5558 and adopts it
5. The daemon's config signature check does not properly detect the passages file change
6. DiskANN C++ sends node IDs from index B to the daemon, but the daemon still has index A's passages loaded
7. The daemon either can't find the node IDs or returns wrong data, causing protobuf parsing to fail

## Reproduction

```bash
# This passes:
pytest tests/test_readme_examples.py::test_readme_basic_example -k diskann -v

# Kill all python processes, then:
# This also passes:
pytest tests/test_readme_examples.py::test_llm_config_simulated -k diskann -v

# But running them together fails on the second test:
pytest tests/test_readme_examples.py -k "readme_basic_example or llm_config_simulated" -k diskann -v
# test_readme_basic_example[diskann] PASSED
# test_llm_config_simulated[diskann] FAILED - RuntimeError: Failed to fetch embeddings
```

Environment: Observed on Windows, but the daemon reuse logic is platform-independent.

## Affected Code

- `packages/leann-core/src/leann/embedding_server_manager.py` - `_adopt_registered_server()` and `_build_config_signature()`: daemon config matching does not properly invalidate when passages content changes (same model + same mode = match, even if passages file is different)

## Possible Fixes

**Option A**: Include passages file path (resolved) in the daemon config signature comparison, so a different passages file triggers a new server start instead of reuse.

**Option B**: Include a hash or mtime of the passages file in the config signature, so content changes are detected even if the path format differs.

**Option C**: On `LeannSearcher.cleanup()`, also terminate the daemon process (trades off daemon reuse performance for correctness in test scenarios).

## Impact

- **Severity**: Medium (causes test failures in sequential runs; could also affect production workflows that build multiple indices in one process)
- **Scope**: Any scenario where multiple DiskANN indices with different data are built and searched in the same process lifetime with daemon mode enabled
- **Workaround**: Kill all Python processes between test runs, or run affected tests in isolation

octo-patch:
Fixes #281

## Problem

When multiple DiskANN tests run sequentially in the same process, a daemonized embedding server started by test A could be incorrectly adopted by test B if the config-signature comparison did not include the passages file path and stat information. This caused the server to serve node IDs from index B while having index A's passages loaded, resulting in:

```
ZMQ_FETCH_ERROR: Server response has invalid dimensions size.
RuntimeError: Failed to fetch embeddings
```

## Solution

Added two regression tests to `test_embedding_server_manager.py` that directly exercise this scenario using the existing mock infrastructure (no real subprocesses or network):

- **`test_different_passages_files_start_separate_daemons`**: Two managers pointing to different passages files (simulating sequential pytest runs each in their own temp directory) must each spawn a separate daemon and use distinct ZMQ ports. Without `passages_file`/`passages_signature` in the config-signature hash key, both managers would map to the same registry file and the second would silently adopt the first daemon.

- **`test_same_passages_file_reuses_existing_daemon`**: The positive case — two managers for the *same* passages file should share the existing daemon (no duplicate spawning).

## Testing

```bash
pytest tests/test_embedding_server_manager.py -v
# 13 passed
```

### DIFF
--- a/packages/leann-core/src/leann/registry.py
+++ b/packages/leann-core/src/leann/registry.py
@@ -63,12 +63,10 @@ def register_project_directory(project_dir: Optional[Union[str, Path]] = None):
     else:
         project_dir = Path(project_dir)
 
-    # Only register directories that have some kind of LEANN content
-    # Either .leann/indexes/ (CLI format) or *.leann.meta.json files (apps format)
+    # Only register directories that have some kind of LEANN content.
+    # Check CLI-format first to avoid an expensive rglob on large directories.
     has_cli_indexes = (project_dir / ".leann" / "indexes").exists()
-    has_app_indexes = any(project_dir.rglob("*.leann.meta.json"))
-
-    if not (has_cli_indexes or has_app_indexes):
+    if not has_cli_indexes and not any(project_dir.rglob("*.leann.meta.json")):
         # Don't register if there are no LEANN indexes
         return
 
--- a/tests/test_embedding_server_manager.py
+++ b/tests/test_embedding_server_manager.py
@@ -489,6 +489,141 @@ def test_registry_record_write_is_atomic(tmp_path, monkeypatch):
     assert dst.endswith(".json")
 
 
+def test_different_passages_files_start_separate_daemons(tmp_path, monkeypatch):
+    """Two managers with different passages files must NOT share a daemon (issue #281).
+
+    Simulates sequential test runs where each test builds its own index in a
+    separate directory but uses the same model and embedding mode.  A stale
+    daemon from run A should never be adopted by run B.
+    """
+    registry_dir = tmp_path / "servers"
+    registry_dir.mkdir()
+    monkeypatch.setattr(EmbeddingServerManager, "_registry_dir", staticmethod(lambda: registry_dir))
+
+    # Each call to _get_available_port returns a strictly increasing port so
+    # that the two daemons land on different ports when they are not sharing.
+    port_seq = [6200]
+
+    def fake_get_available_port(start_port):
+        p = port_seq[0]
+        port_seq[0] += 1
+        return p
+
+    monkeypatch.setattr(
+        "leann.embedding_server_manager._get_available_port",
+        fake_get_available_port,
+    )
+
+    alive_pids: set[int] = set()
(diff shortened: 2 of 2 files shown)


ASuresh0524:
Thanks for the regression coverage for #281 

A few small things that would make this even easier to merge/review:

- PR description/title — The branch also includes a tiny register_project_directory change: skip the *.leann.meta.json rglob when .leann/indexes already exists. That’s unrelated to daemon signatures but reasonable to ship together; worth one sentence in the PR body so reviewers don’t wonder why registry.py is in the same PR (or split into a follow-up if maintainers prefer strict scope).
- Squash / merge commit message - test: regression for daemon isolation by passages (#281); perf: skip meta rglob when CLI indexes exist so release notes reflect both bullets.
- Tests — Optional: a one-line comment above passages_file=str(meta_a) noting that passages_file is the index meta path (same as other tests in this file) avoids future confusion with the literal .passages.jsonl path.

CI is green; from here this looks good to merge once the description calls out the registry tweak (or it’s split).

PR #306, opened 2026-04-27
feat(cli): add --metadata-filters flag to leann search

### LINKED ISSUE #85: Metadata Filtering
Hello,

I've been trying for some time to filter indexed files based on their metadata added during the build.

For example, in the example given in the documentation :

```
metadata_filters={
"file_extension": {"==": ".py"},
"lines_of_code": {"<": 100}
}
)
```

It doesn't matter whether I set it or not. Even if I set false conditions, it still returns the result.

Did I miss something?

Thank you.

octo-patch:
Fixes #85

## Problem

The Python API has supported `metadata_filters` in `LeannSearcher.search()` for some time, but `leann search` (the CLI) never exposed this parameter. Users who wanted to filter results by metadata were forced to drop down to the Python API — the CLI silently ignored any filtering intent.

From issue #85: *"Even if I set false conditions, it still returns the result"* — this was partly a Python API issue (since fixed) and partly because there was no CLI path to reach the filtering at all.

## Solution

Add `--metadata-filters` to `leann search`, accepting a JSON string with the same format as the Python API:

```bash
# Filter to chapters 1–5 in the fiction genre
leann search alice-index "what happens to Alice" \
  --metadata-filters '{"chapter": {"<=": 5}, "genre": {"==": "fiction"}}'

# Combine with --json for machine-readable filtered output
leann search my-docs "database" \
  --metadata-filters '{"file_extension": {"==": ".py"}}' \
  --json
```

Supported operators (same set as `MetadataFilterEngine`): `==`, `!=`, `<`, `<=`, `>`, `>=`, `in`, `not_in`, `contains`, `starts_with`, `ends_with`.

Invalid JSON and non-object values produce a clear error message before any search is attempted, instead of crashing.

## Testing

- Added `TestCLIMetadataFilterParsing` to `tests/test_metadata_filtering.py`:
  - Valid JSON parsing (equality, numeric range, compound)
  - `None` / empty string → `None` (no filter)
  - Invalid JSON → clear error string
  - Non-dict JSON (array) → clear error string
  - Round-trip: parsed filters produce correct results via `MetadataFilterEngine`

### DIFF
--- a/packages/leann-core/src/leann/cli.py
+++ b/packages/leann-core/src/leann/cli.py
@@ -436,9 +436,9 @@ Examples:
             type=str,
             default=None,
             help=(
-                'Filter results by metadata fields (JSON string). '
+                "Filter results by metadata fields (JSON string). "
                 'Format: \'{"field": {"operator": value}}\'. '
-                'Operators: ==, !=, <, <=, >, >=, in, not_in, contains, starts_with, ends_with. '
+                "Operators: ==, !=, <, <=, >, >=, in, not_in, contains, starts_with, ends_with. "
                 'Example: \'{"chapter": {"<=": 5}, "genre": {"==": "fiction"}}\''
             ),
         )
--- a/packages/leann-core/src/leann/cli.py
+++ b/packages/leann-core/src/leann/cli.py
@@ -431,6 +431,17 @@ Examples:
             default=True,
             help="Enable/disable warmup when starting embedding server (default: enabled)",
         )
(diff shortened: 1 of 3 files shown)


ASuresh0524:
Please fix CI

octo-patch:
Thanks for the heads up. Pushed a fix in 419a440 — applied `ruff format` to `cli.py` and bumped `uv.lock` revision from 1 to 3 to satisfy the lint hook. CI should be green now.

PR #314, opened 2026-05-17
feat(browser_rag): put url/domain/visit fields in metadata so they're filterable

natekettles:
closed - sorry for noise.


natekettles:
Re-opened as #315 — sorry for the noise, had to delete and re-fork to clean up my fork's main branch. Same code, same description.

PR #338, opened 2026-05-20
[content-hash 5/5] feat: switch default passage_id_scheme to content-hash

raoabinav:
## Evaluator Summary

- Abi changed the default passage ID scheme for new LEANN indexes to content hashes after the metadata, builder, reader, and migration foundations were in place.
- The design is backward-compatible: existing indexes continue using the scheme recorded in their metadata, and users can explicitly opt back into sequential IDs.
- The implementation updates builder defaults, CLI defaults, and incremental rebuild helpers so new indexes get stable IDs without breaking old artifacts.
- Quality bar: covered by the content-hash stack tests for defaults, CLI behavior, and existing-index compatibility.

### DIFF
--- a/packages/leann-core/src/leann/api.py
+++ b/packages/leann-core/src/leann/api.py
@@ -497,23 +497,37 @@ class LeannBuilder:
         self.backend_kwargs = backend_kwargs
         self.chunks: list[dict[str, Any]] = []
 
-    def _generate_passage_id(self, text: str) -> str:
+    @staticmethod
+    def _make_unique_passage_id(base_id: str, reserved_ids: set[str]) -> str:
+        if base_id not in reserved_ids:
+            return base_id
+        suffix = 1
+        while f"{base_id}-{suffix}" in reserved_ids:
+            suffix += 1
+        return f"{base_id}-{suffix}"
+
+    def _generate_passage_id(self, text: str, reserved_ids: Optional[set[str]] = None) -> str:
         """Generate a passage ID per the configured scheme.
 
         sequential: str(insertion index) — fast, position-dependent, current default.
-        content-hash: sha256(text)[:16] — content-stable, dedup-friendly across
-        file moves and reorderings. See #329 for the design.
+        content-hash: sha256(text)[:16] — content-stable across file moves and
+        reorderings when text is unique. Duplicate text receives a numeric suffix
+        so the passage store, ID maps, and vector labels stay one-to-one.
         """
         if self.passage_id_scheme == PASSAGE_ID_SCHEME_CONTENT_HASH:
             import hashlib
 
-            return hashlib.sha256(text.encode("utf-8")).hexdigest()[:16]
+            base_id = hashlib.sha256(text.encode("utf-8")).hexdigest()[:16]
+            return self._make_unique_passage_id(base_id, reserved_ids or set())
         return str(len(self.chunks))
 
     def add_text(self, text: str, metadata: Optional[dict[str, Any]] = None):
         if metadata is None:
             metadata = {}
-        passage_id = metadata.get("id") or self._generate_passage_id(text)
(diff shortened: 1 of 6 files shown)


raoabinav:
Keeping this separate from #347. #347 adds opt-in content-hash IDs and migration support; this PR changes the default and should only be reconsidered after the opt-in path lands.

packages/leann-core/src/leann/api.py
```diff
@@ -881,14 +936,28 @@ def update_index(self, index_path: str, remove_passage_ids: Optional[list[str]]
...
         for chunk in self.chunks:
             text = chunk.get("text", "")
             if not isinstance(text, str) or not text.strip():
                 continue
             metadata = chunk.setdefault("metadata", {})
-            passage_id = chunk.get("id") or metadata.get("id")
-            if passage_id and passage_id in existing_ids:
+            if "id" in metadata and metadata["id"] is not None:
+                passage_id = str(metadata["id"])
+            elif self.passage_id_scheme == PASSAGE_ID_SCHEME_CONTENT_HASH:
```
jane-jhu, line 948: `update_index()` should derive the effective passage ID scheme from the existing index metadata, not from the builder default. Direct API callers can now create `LeannBuilder(backend_name="hnsw")`, which defaults to content-hash, and update a legacy sequential index, producing mixed IDs. The CLI guards this via `_make_incremental_builder()`, but the public API path does not.

packages/leann-core/src/leann/api.py
```diff
@@ -714,12 +768,13 @@ def build_index_from_arrays(self, index_path: str, ids: list, embeddings: np.nda
...
         leann_meta_path = index_dir / f"{index_name}.meta.json"
         meta_data = {
-            "version": "1.0",
+            "version": "1.1",
             "backend_name": self.backend_name,
             "embedding_model": self.embedding_model,
             "dimensions": self.dimensions,
             "backend_kwargs": self.backend_kwargs,
             "embedding_mode": self.embedding_mode,
+            "passage_id_scheme": self.passage_id_scheme,
```
jane-jhu, line 777: In this path, `self.passage_id_scheme` may be the backend default (`content-hash` for HNSW/IVF), but `build_index_from_arrays()` also accepts explicit `ids` and writes those IDs into the backend/id map. If callers pass IDs like `"0"`/`"1"`, the metadata can say `content-hash` even though the actual stored IDs are caller-provided. Should this path either force an explicit/external scheme or make the passage store and id map consistently use generated content-hash IDs?

packages/leann-core/src/leann/cli.py
```diff
@@ -2384,20 +2555,40 @@ async def build_index(self, args):
...
+            for fs in target_synchronizers:
+                fs.create_snapshot()
+            self._write_sync_config(
+                target_index_dir,
+                self._resolve_sync_roots(docs_paths),
+                self._parse_file_types(args.file_types),
+                self._sync_ignore_patterns(args.include_hidden),
+            )
+            if publish_from_staging:
+                _publish_rebuilt_index(target_index_dir, index_dir)
```
jane-jhu, line 2585: The new staging publish flow protects against rebuild failures, but it does not acquire `index_write_lock`. Since `migrate_ids()` uses that lock, a concurrent `build --force` can still publish/replace the same index while migration is rewriting sidecars. Should the rebuild publish path also take `index_write_lock(index_dir)`?

jane-jhu (changes requested):

raoabinav:
Parking this default-flip PR for now. #347 is the merge-ready opt-in content-hash path; this PR changes the default and has valid follow-up concerns around direct API incremental updates, array-built indexes with caller-provided IDs, and rebuild publish locking. I’ll revisit this as a smaller clean PR after #347 lands.

PR #340, opened 2026-05-25
Fix MPS memory pathologies on Apple Silicon

ww2283:
## Problem

Two issues in `embedding_compute.py` cause 22 GB GPU memory usage for a 110M BERT model on Apple Silicon (expected ~3 GB):

1. `torch.mps.set_per_process_memory_fraction(0.9)` tells MPS allocator to use up to 29 GB on a 32 GB machine. Allocator fills budget and never releases.
2. `torch.compile(mode="reduce-overhead", dynamic=True)` on MPS caches compiled graphs + pre-allocated FP16 buffers per sequence-length bucket. 640 batches of varying lengths create 5+ GB of graph buffers.

Additionally, `BM25Scorer` has a syntax error from the bm25 refactor merge — stray unquoted text between `__init__` and `Fts5BM25Index` breaks `import leann`.

## Footprint data (sudo footprint, 32 GB M-series Mac)

| Configuration | phys_footprint |
|---|---|
| Original (both issues) | 22 GB |
| torch.compile removed only | 17 GB |
| Both fixed + empty_cache() | ~3 GB |

Model: lokeshch19/ModernPubMedBERT (~110M params), batch_size=8, 5114 chunks, 640 batches.

## Fix

- Remove `set_per_process_memory_fraction` for MPS (no equivalent benefit to CUDA's OOM protection)
- Guard `torch.compile` to `device == "cuda"` only (2 call sites)
- Add `torch.mps.empty_cache()` between manual HF tokenization batches
- Restore `BM25Scorer.fit()` and `.search()` methods lost in bm25 refactor series

### DIFF
--- a/packages/leann-core/src/leann/api.py
+++ b/packages/leann-core/src/leann/api.py
@@ -298,8 +298,6 @@ class BM25Index(ABC):
         """
 
 
-
-
 class Fts5BM25Index(BM25Index):
     """BM25 over a SQLite FTS5 virtual table, persisted on disk.
 
@@ -381,9 +379,7 @@ class LeannBuilder:
         **backend_kwargs,
     ):
         if bm25_backend != "fts5":
-            logger.warning(
-                f"bm25_backend={bm25_backend!r} is deprecated; using 'fts5'."
-            )
+            logger.warning(f"bm25_backend={bm25_backend!r} is deprecated; using 'fts5'.")
             bm25_backend = "fts5"
         self.bm25_backend = bm25_backend
         self.prebuild_bm25 = prebuild_bm25 or bm25_backend == "fts5"
@@ -623,7 +619,6 @@ class LeannBuilder:
         index.fit(self.chunks)
         index.close()
         logger.info(f"Wrote BM25 FTS5 index to {db_path}")
-
 
     def build_index_from_arrays(self, index_path: str, ids: list, embeddings: np.ndarray):
         """Build an index from pre-computed embedding arrays.
--- a/packages/leann-core/src/leann/api.py
+++ b/packages/leann-core/src/leann/api.py
@@ -1462,11 +1462,33 @@ class LeannSearcher:
         index = Fts5BM25Index(str(db_path))
         passages = []
         for passage_file in self.passage_manager.passage_files.values():
-            with open(passage_file, encoding="utf-8") as f:
-                for line in f:
-                    if line.strip():
-                        passages.append(json.loads(line))
-        index.fit(passages)
+            try:
+                with open(passage_file, encoding="utf-8") as f:
+                    for line in f:
+                        if line.strip():
+                            try:
+                                passages.append(json.loads(line))
+                            except json.JSONDecodeError as exc:
+                                logger.warning(f"Skipping malformed JSONL in {passage_file}: {exc}")
+            except FileNotFoundError:
+                logger.warning(f"Passage file missing: {passage_file}")
+
+        if not passages:
+            logger.error(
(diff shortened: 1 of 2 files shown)


yichuan-w:
This is such a nice PR, we should merge that cc @andylizf @ASuresh0524

andylizf (commented):
LGTM — MPS data is solid, BM25 cleanup is clean. Two nits below.

packages/leann-core/src/leann/api.py
```diff
@@ -1585,31 +1452,40 @@ def _init_bm25(self) -> None:
...
+            except FileNotFoundError:
+                logger.warning(f"Passage file missing: {passage_file}")
+
+        if not passages:
+            logger.error(
+                "No passages found for on-demand BM25 index. "
+                "BM25/hybrid search will return empty results. "
+                "Re-run 'leann build' to regenerate passage files."
+            )
+
```
andylizf, line 1478: nit: After logging the error, this falls through to `index.fit([])` which creates an empty FTS5 table. Consider adding a `return` here to bail out early — an empty index silently returns no results on search, which is harder to debug than skipping BM25 init entirely.

packages/leann-core/src/leann/api.py
```diff
@@ -1585,31 +1452,40 @@ def _init_bm25(self) -> None:
...
+        try:
+            index.fit(passages)
+        except (PermissionError, OSError) as exc:
+            logger.error(
+                f"Cannot write BM25 index to {db_path}: {exc}. "
+                f"Ensure the index directory is writable, or rebuild with prebuild_bm25=True."
+            )
+            return
+
```
andylizf, line 1487: nit: When `fit()` raises PermissionError/OSError, `self.bm25_scorer` is never assigned. Worth confirming the caller guards against `bm25_scorer` being unset — otherwise `_bm25_search` will AttributeError on the next hybrid query.

yichuan-w:
yeah, and please fix the conflict here

packages/leann-core/src/leann/api.py
ww2283: Good catch — fixed in f9c74aa. The empty-passages path now `return`s early, leaving `self.bm25_scorer = None`. `_bm25_search` (line 1496-1497) checks for that and raises `RuntimeError("BM25 scorer failed to initialize")`, so it surfaces loudly instead of returning empty hits.

packages/leann-core/src/leann/api.py
ww2283: Confirmed safe — `_bm25_search` at lines 1496-1497 already does `if scorer is None: raise RuntimeError("BM25 scorer failed to initialize")`. The PermissionError/OSError path `return`s without assigning `self.bm25_scorer`, so it stays `None` and the next call surfaces as RuntimeError, not AttributeError. With f9c74aa, the empty-passages path now behaves the same way.

ww2283:
Rebased onto main (post-#341) and pushed f9c74aa. Conflict resolution: kept the deletion path (this PR completes the #335 intent that PR #341's author noted in their description). Both review nits addressed:

- **nit 1** (empty passages): now `return`s early; empty FTS5 table no longer silently swallows queries.
- **nit 2** (PermissionError/OSError): confirmed the existing `_bm25_search` guard at L1496-1497 converts `bm25_scorer is None` into `RuntimeError`, so no AttributeError reaches callers.

Credential audit on the 5 commits in this PR: clean — only `api.py` and `embedding_compute.py` touched, no secrets.

PR #347, opened 2026-06-02
feat: support content-hash passage IDs

raoabinav:
- added content-hash passage IDs so passage identity is stable across file moves, reorderings, and rebuilds instead of depending on sequential insertion order.
- The design records the selected passage ID scheme in metadata and keeps incremental updates aligned with the existing index scheme to avoid mixed-ID indexes.
- The implementation includes migration support for existing sequential-ID indexes and preserves compatibility with older indexes.
- Quality bar: tests cover content-hash generation, metadata persistence, update compatibility, migration behavior, and duplicate/collision handling.

### DIFF
--- a/packages/leann-core/src/leann/cli.py
+++ b/packages/leann-core/src/leann/cli.py
@@ -16,7 +16,7 @@ from llama_index.core import SimpleDirectoryReader
 from llama_index.core.node_parser import SentenceSplitter
 from tqdm import tqdm
 
-from .api import LeannBuilder, LeannChat, LeannSearcher
+from .api import Fts5BM25Index, LeannBuilder, LeannChat, LeannSearcher
 from .embedding_server_manager import EmbeddingServerManager
 from .interactive_utils import create_cli_session
 from .registry import register_project_directory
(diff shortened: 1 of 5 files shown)


packages/leann-core/src/leann/cli.py
```diff
@@ -1880,7 +1910,32 @@ def _chunks_for_paths(self, all_texts: list[dict], paths: set[str]) -> list[dict
...
+        "fall back to whatever the args say or the default".
+        """
+        meta_path = Path(index_path).with_suffix(".leann.meta.json")
+        if not meta_path.exists():
+            return None
+        try:
+            with open(meta_path, encoding="utf-8") as f:
+                return json.load(f).get("passage_id_scheme")
+        except Exception:
+            return None
```
jane-jhu, line 2026: For an existing meta file that lacks passage_id_scheme, we should treat it as sequential rather than returning None. Otherwise legacy indexes can fall back to args.id_scheme/current default during incremental updates, which will become dangerous once the default flips to content-hash. Consider returning "sequential" when the meta exists but the field is absent.

raoabinav: Good suggestion, thanks. I was thinking of `None` as the “no existing index / no recorded scheme” path, but needs to be different

jane-jhu (approved):

PR #348, opened 2026-06-02
feat(backend): add GPU FlashLib IVF backend (flashlib_ivf) + IVF-vs-I…

andy-yang-1:
…VF benchmark

Add leann-backend-flashlib-ivf, a GPU IVF-Flat (inverted file) approximate-NN backend built on FlashLib's flash_ivf_flat (Triton/CuteDSL) - the GPU counterpart of the FAISS `ivf` backend, sharing its nlist/nprobe recall knobs so the two are drop-in comparable. The built index (centroids/data/ids/CSR offsets) is persisted with torch.save and reloaded to the GPU at search time (no k-means re-train). mips/cosine L2-normalize (FlashLib IVF ranks by squared L2).

Also add benchmarks/flashlib_ivf_vs_faiss_ivf.py (flashlib_ivf GPU vs ivf CPU at a matched nlist across an nprobe sweep: build, latency, throughput, recall@k vs exact GT), a CUDA-guarded correctness test, the flashlib-ivf extra + uv source wiring, and a flashlib_ivf section in the backend guide.

On an H200 at 1M x 768 (nlist=4096, 8 CPU threads, cosine): ~13x faster build and, at nprobe=32, ~6.5x lower single-query latency / ~75x higher batched throughput at comparable recall (GPU latency ~flat vs CPU linear in nprobe).

## Related Issues

Fixes #

## Checklist

### DIFF
--- a/benchmarks/flashlib_ivf_vs_faiss_ivf.py
+++ b/benchmarks/flashlib_ivf_vs_faiss_ivf.py
@@ -1,0 +1,411 @@
+#!/usr/bin/env python3
+"""
+FlashLib IVF (GPU) vs FAISS IVF (CPU) head-to-head for LEANN.
+
+This is the apples-to-apples *approximate* comparison: both backends are IVF-Flat
+(inverted file) indexes that coarse-quantize the corpus into ``nlist`` cells and, at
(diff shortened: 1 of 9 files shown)


yichuan-w:
Nice PR to support a user who has an advanced GPU w/o recompute, nice work @andy-yang-1

PR #349, opened 2026-06-03
feat: add Obsidian vault ingestion

raoabinav:
## Evaluator Summary

- Abi added an Obsidian vault ingestion example that indexes Markdown notes while preserving vault-aware metadata such as relative paths, titles, aliases, tags, wikilinks, embeds, and frontmatter.
- The design treats Obsidian as a graph-shaped local knowledge base, so retrieval results carry enough metadata for backlink-aware and note-aware follow-ups.
- The implementation is scoped to Markdown vault ingestion rather than an Obsidian plugin or live sync, keeping the first PR reviewable and testable.
- Quality bar: focused tests cover vault parsing and metadata extraction; lint/format checks were run for the app and tests.

### DIFF
--- a/apps/obsidian_rag.py
+++ b/apps/obsidian_rag.py
@@ -1,0 +1,251 @@
+"""
+Obsidian vault RAG example.
+
+Indexes Markdown notes from an Obsidian vault while preserving note metadata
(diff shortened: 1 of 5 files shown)


raoabinav:
Replaced by #358 from the neutral feature branch.

PR #353, opened 2026-06-03
feat: add code context metadata foundation

raoabinav:
## Evaluator Summary

- Abi added an AST-aware code-context metadata layer for Python chunks, capturing module paths, qualified symbols, definitions, imports, calls, and references so LEANN can reason over code structure instead of only raw text.
- The design keeps extraction pure-Python and metadata-only in this PR: low-risk, serializable, and suitable as a stable contract for later related-code expansion and cross-file symbol traversal.
- The implementation integrates at chunking time, so downstream search, MCP, and agent workflows can consume richer code metadata without changing the vector backend API.
- Quality bar: focused tests cover code-context extraction, AST chunk integration, RAG app no-recompute behavior, and chat propagation; lint/format checks were run on the changed modules.

### DIFF
--- a/apps/base_rag_example.py
+++ b/apps/base_rag_example.py
@@ -359,6 +359,7 @@ class BaseRAGExample(ABC):
                 query,
                 top_k=args.top_k,
                 complexity=args.search_complexity,
+                recompute_embeddings=not args.no_recompute,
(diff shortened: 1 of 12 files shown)


raoabinav:
Replaced by #354 from the neutral feature branch.

PR #359, opened 2026-06-04
feat: add indexed regex search

raoabinav:
## Evaluator Summary

- Abi added indexed exact search through a local SQLite trigram sidecar built next to LEANN's passage and vector index files, giving agents fast literal and regex search without scanning every passage.
- Design choice: trigrams are only a candidate filter; final matching still uses deterministic Python regex verification, and unsafe regex patterns fall back to live passage scanning to avoid false negatives.
- The feature integrates with build, update, lazy old-index sidecar creation, API search, chat retrieval, CLI search, and the HTTP server while preserving `use_grep=True` as a portable literal-search compatibility path.
- Quality bar: 5 targeted tests cover literal search, regex search, invalid regex handling, metadata filters, and lazy sidecar creation, plus lint, format, type, CLI help, lockfile, and whitespace checks.


raoabinav:

PR #360, opened 2026-06-04
feat: add Local Cursor proxy

raoabinav:
## Evaluator Summary

- Abi added `leann cursor`, a loopback-by-default OpenAI-compatible proxy that lets local coding assistants retrieve LEANN code context before forwarding requests to Ollama, LM Studio, or another local model server.
- Design choice: the proxy keeps the client protocol standard (`/v1/models` and `/v1/chat/completions`) while injecting retrieved snippets into the system prompt, so editors can use it without a custom integration.
- The implementation includes streaming and non-streaming passthrough, explicit CORS allowlists, request-size limits, JSON validation, configurable retrieval depth, and opt-out embedding recomputation.
- Quality bar: 10 targeted tests cover proxy validation, retrieval injection, streaming behavior, and local-server safety, plus lint, format, type, lockfile, and whitespace checks.


raoabinav:

PR #361, opened 2026-06-04
feat: harden ReAct source policy

raoabinav:
## Evaluator Summary

- Abi added source-policy controls to LEANN's ReAct agent so runs can be explicitly local-only, web-only, or combined, with unavailable tools hidden from prompts and refused at runtime.
- Design choice: LEANN now has provider-neutral research-tool contracts for local search, web search, and page fetches, plus optional Claude/OpenAI agent SDK adapter helpers that do not affect the default install.
- The web path adds bounded Serper search and Jina Reader fetches with timeouts, HTTP(S)-only URL validation, clear missing-key behavior, and runtime refusal for tools disabled by source policy.
- Quality bar: 30 targeted tests cover tool registry behavior, optional SDK wrappers, routing, missing API keys, web failures, and source-policy enforcement, plus lint, format, and type checks on the touched agent files.


raoabinav:

PR #363, opened 2026-06-04
feat: add opt-in OCR PDF indexing

raoabinav:
## Evaluator Summary

- Abi added opt-in OCR indexing for scanned and image-heavy PDFs, so LEANN can retrieve content from documents that normal embedded-text extraction would miss.
- Design choice: OCR is only enabled through `leann build --enable-ocr` and only runs on pages without embedded text, preserving the fast default path for ordinary PDFs.
- The implementation covers directory-discovered PDFs and explicit PDF file inputs, adds an OCR example app, documents the optional dependency and Tesseract binary requirement, and keeps OCR behavior isolated from normal text ingestion.
- Quality bar: 7 targeted tests cover blank-page OCR, text-page skipping, explicit-file handling, and dependency stubs, plus lint, format, type, lockfile, and whitespace checks.

### DIFF
--- a/apps/ocr_rag.py
+++ b/apps/ocr_rag.py
@@ -1,0 +1,102 @@
+"""
+OCR RAG example for scanned and image-heavy PDFs.
+
+This app extracts embedded PDF text first, then OCRs only pages without text.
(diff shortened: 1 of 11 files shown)


raoabinav:

PR #364, opened 2026-06-04
feat: harden CLIP image RAG

raoabinav:
## Evaluator Summary

- Abi hardened LEANN's CLIP image retrieval path so local image corpora can be indexed with deterministic discovery, stable IDs, rich metadata, and text-to-image search.
- Design choice: image passages use stable `image-<hash>` identifiers and store file/path/size/dimension/model metadata, which makes rebuilds reproducible and result inspection reviewer-friendly.
- The implementation builds from precomputed CLIP embedding arrays and disables query-time recomputation for image vectors, because stored text cannot recreate the original pixels.
- Quality bar: 6 targeted fake-encoder tests avoid model downloads while covering discovery, metadata, stable IDs, and retrieval wiring, plus lint, format, type, lockfile, and whitespace checks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -443,6 +443,18 @@ python -m apps.code_rag --repo-dir "./my_codebase" --query "How does authenticat
 
 </details>
 
+### 🖼️ CLIP Image Search: Find Images with Text
+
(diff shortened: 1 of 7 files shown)


raoabinav:

PR #365, opened 2026-06-04
feat: add video frame retrieval

raoabinav:
## Evaluator Summary

- Abi added frame-level video retrieval by sampling local video frames, embedding them with CLIP, and indexing each sampled frame as a searchable LEANN passage.
- Design choice: each frame gets a stable `video-frame-<hash>` ID plus timestamp, frame index, frame size, duration, source path, media type, and model metadata, so search results point back into the original video.
- The implementation builds from precomputed CLIP frame arrays and disables query-time recomputation for frame vectors, matching the image-retrieval design while keeping visual retrieval separate from audio transcription.
- Quality bar: 9 targeted tests mock decoding and model inference to avoid codec/model downloads while covering sampling, metadata, stable IDs, and retrieval wiring, plus lint, format, type, lockfile, and whitespace checks.

### DIFF
--- a/.github/workflows/build-reusable.yml
+++ b/.github/workflows/build-reusable.yml
@@ -659,14 +659,36 @@ jobs:
         run: |
           source .venv/bin/activate || source .venv/Scripts/activate
           python - <<'PY'
+          import numpy as np
           import leann
           import leann_backend_hnsw as h
           import leann_backend_diskann as d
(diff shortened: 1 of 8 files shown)


yichuan-w:
I don't think CLIP is a very good embedding model to use nowadays, especially for video, maybe Qwen3VL embedding is better

raoabinav:

PR #366, opened 2026-06-04
fix: polish native Windows paths

raoabinav:
## Evaluator Summary

- Abi fixed native Windows portability issues in LEANN's chat and local-model integration paths, improving behavior outside Unix-like shells.
- Design choice: the patch stays narrow and uses platform-native abstractions, replacing POSIX-only assumptions with `os.pathsep` for `NODE_PATH` and guarded timeout behavior where `SIGALRM` is unavailable.
- The implementation also makes the CPU `mkldnn.enabled` setting type-checker-friendly, reducing platform-specific lint/type noise in embedding setup.
- Quality bar: focused Windows and LM Studio tests passed locally, plus lint, format, type, lockfile, and whitespace checks; this is a clean supporting PR rather than a flagship roadmap feature.

### DIFF
--- a/packages/leann-core/src/leann/chat.py
+++ b/packages/leann-core/src/leann/chat.py
@@ -8,6 +8,7 @@ import difflib
 import logging
 import os
 from abc import ABC, abstractmethod
+from collections.abc import Callable
 from typing import Any, Optional, cast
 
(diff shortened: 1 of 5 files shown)


raoabinav:

PR #367, opened 2026-06-05
ci: harden uv dependency installs

raoabinav:
## Summary
- Adds uv HTTP retry/timeout settings for the large build matrix so transient PyPI/CDN stalls do not fail PR runs immediately.
- Enables setup-uv cache for lint, type-check, and wheel-build jobs using uv.lock as the cache dependency key.
- Wraps uv build/install calls that touch remote dependencies with bounded retries.
- Disables matrix fail-fast so one infrastructure timeout does not cancel the remaining platform signal.

## Context
Recent PR runs failed on macOS Intel before tests executed because uv timed out while downloading PyPI wheels such as pandocfilters and jupyter-core. This PR hardens the workflow against that failure mode without changing package code.

## Validation
- Parsed .github/workflows/build-reusable.yml locally with Ruby YAML loader.
- Ran git diff --check.

### DIFF
--- a/.github/workflows/build-reusable.yml
+++ b/.github/workflows/build-reusable.yml
@@ -9,6 +9,15 @@ on:
         type: string
         default: ''
 
+env:
+  # Large matrix builds download many Python wheels in parallel. Give transient
+  # PyPI/CDN stalls enough room to recover instead of failing a whole PR run.
+  UV_HTTP_RETRIES: "8"
+  UV_HTTP_TIMEOUT: "120"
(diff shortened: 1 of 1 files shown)


raoabinav:

raoabinav:
Reopening this now that #347 has landed. This is intentionally scoped to CI reliability only: uv retries/timeouts, setup-uv cache, bounded retries around uv build/install steps, and fail-fast disabled so one transient download failure does not cancel the rest of the matrix.

PR #368, opened 2026-06-06
fix(embeddings): make CUDA batch size configurable for small GPUs

### LINKED ISSUE #344: CUDA embedding batch size hardcoded to 256 — OOMs on small GPUs with no override
**Version:** leann-core 0.3.7
**Env:** NVIDIA RTX A1000 (4 GB VRAM), WSL2, Python 3.12, torch 2.12 (CUDA)

### Problem
In `leann/embedding_compute.py`, `compute_embeddings_sentence_transformers()` hardcodes the CUDA batch size when `adaptive_optimization=True`:

```python
elif device == "cuda":
    batch_size = 256  # CUDA optimal batch size
```

A 256-wide batch of up-to-512-token sequences won't fit a 4 GB GPU (or really anything under ~8 GB), so `leann build` dies with `torch.OutOfMemoryError: CUDA out of memory` during the embedding pass. There is a `provider_options["batch_size"]` override path in `compute_embeddings()`, but it isn't reachable from the `build`/`watch` CLI, so the only workaround is editing the installed source.

A related limitation on the CPU path caps threads at `min(8, os.cpu_count())` (`embedding_compute.py`, `device == "cpu"`), which underutilizes many-core machines (I observed only ~3–8 cores used on a 16-core box, making the CPU fallback very slow for a few thousand chunks).

### Repro
On a ≤4 GB CUDA GPU:

```
leann build demo --docs ./src \
  --embedding-mode sentence-transformers --embedding-model BAAI/bge-base-en-v1.5
```

→ OOM during embedding (a few thousand chunks). The model loads fine and a tiny corpus (single small dir) can sneak through with one small batch, but a realistic codebase OOMs.

### Expected
Builds succeed on small GPUs, or there is a documented knob to size the batch.

### Suggested fix
1. Make the CUDA batch size overridable, e.g. `batch_size = int(os.getenv("LEANN_CUDA_BATCH_SIZE", "256"))` — keeps the current default, lets small-GPU users opt down. Same idea for the CPU thread cap.
2. Better: auto-scale to free VRAM via `torch.cuda.mem_get_info`, or catch CUDA OOM and retry with a halved batch.

Happy to open a PR for option 1 (plus a VRAM-aware variant) if that's welcome.

ASuresh0524:
Fixes #344. LEANN_CUDA_BATCH_SIZE / LEANN_MPS_BATCH_SIZE / LEANN_CPU_THREADS env vars and --embedding-batch-size CLI override hardcoded defaults; cap CUDA batch by free VRAM after model load and halve on OOM retry.

## Checklist

### DIFF
--- a/packages/leann-core/src/leann/cli.py
+++ b/packages/leann-core/src/leann/cli.py
@@ -298,6 +298,12 @@ Examples:
             help="Prompt template to prepend to all texts for embedding (e.g., 'query: ' for search)",
         )
         build_parser.add_argument(
+            "--embedding-batch-size",
+            type=int,
+            default=None,
+            help="Embedding batch size for sentence-transformers (overrides device defaults; also set LEANN_CUDA_BATCH_SIZE env)",
+        )
+        build_parser.add_argument(
             "--query-prompt-template",
             type=str,
             default=None,
@@ -715,6 +721,12 @@ Examples:
                 help="OpenAI-compatible embedding base URL",
             )
             p.add_argument("--embedding-api-key", type=str, default=None, help="Embedding API key")
+            p.add_argument(
+                "--embedding-batch-size",
+                type=int,
+                default=None,
+                help="Embedding batch size for sentence-transformers",
+            )
             p.add_argument(
                 "--max-count", type=int, default=1000, help="Max items to index (default: 1000)"
             )
@@ -1828,6 +1840,9 @@ Examples:
             opts["query_prompt_template"] = args.query_prompt_template
         elif args.embedding_prompt_template:
             opts["prompt_template"] = args.embedding_prompt_template
+        batch_size = getattr(args, "embedding_batch_size", None)
+        if batch_size is not None:
+            opts["batch_size"] = batch_size
         return opts
 
     def _build_config_from_args(
--- a/packages/leann-core/src/leann/embedding_compute.py
+++ b/packages/leann-core/src/leann/embedding_compute.py
@@ -320,6 +320,104 @@ def _query_lmstudio_context_limit(model_name: str, base_url: str) -> Optional[in
 # Global model cache to avoid repeated loading
 _model_cache: dict[str, Any] = {}
 
+_DEFAULT_CUDA_BATCH_SIZE = 256
+_DEFAULT_MPS_BATCH_SIZE = 128
+
+
+def _parse_positive_int_env(name: str, default: int) -> int:
+    raw = os.getenv(name)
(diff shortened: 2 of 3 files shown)


ArtifexSystems:
Tested this on the exact small-GPU setup from #344 and it works — it turns the OOM into a clean build. 👍

**Setup:** NVIDIA RTX A1000 Laptop GPU (4 GB), WSL2, `leann-core` 0.3.7, `torch` 2.12.0+cu130, `sentence-transformers` 5.5.1, model `BAAI/bge-base-en-v1.5`. Ran the PR head (`80e83da`) against a realistic corpus of 384 × 512-token chunks.

**Results**
- **Env / CLI override** — unset → 256 (prior default preserved); `LEANN_CUDA_BATCH_SIZE=24` → 24; an invalid value warns and falls back; `LEANN_CPU_THREADS=16` honored. ✅
- **VRAM cap** — at ~3.2 GiB free, `_cap_cuda_batch_by_vram(256)` → 192; `LEANN_CUDA_AUTO_BATCH=0` opts out → 256. ✅
- **End-to-end default path** (`compute_embeddings(..., is_build=True)`, no env set): auto-capped 256 → 153, that **still OOM'd**, and the halving retry dropped to 76 → **succeeded `(384, 768)`**. A hard `torch.OutOfMemoryError` became a clean build with zero manual tuning. ✅

Two small notes from the run, in case they're useful:

1. **The VRAM heuristic is a touch optimistic for ≤4 GB.** `0.35 × free ÷ ~6 MB/seq` lands around 150–210 on a 4 GB card, which still OOMs here — the halving retry is what actually makes it land (at 76). The retry makes it robust regardless, but a more conservative factor (or sizing against the eager-attention `batch × heads × seq²` peak, which dominates the footprint) would let the common path succeed without leaning on the retry.
2. **The cap runs even when the batch size was set explicitly.** `_cap_cuda_batch_by_vram` applies whenever `device == "cuda"`, so a user who deliberately passes `--embedding-batch-size N` (which sets `adaptive_optimization=False`) still gets it reduced unless they also set `LEANN_CUDA_AUTO_BATCH=0`. Might be friendlier to skip the auto-cap when the value came from an explicit flag.

Neither blocks anything — this is a complete fix for #344: env/CLI override **and** VRAM autoscale + OOM retry, with no new dependency. Once it merges I'll drop our local monkey-patch in favor of these knobs. Thanks @ASuresh0524!

ASuresh0524:
Awesome, thanks @ArtifexSystems. That helps a lot! Will incorporate these two small changes.

PR #377, opened 2026-06-16
fix(watch): byte-hash sync with build extension filter (#345)

### LINKED ISSUE #345: leann watch hard-crashes on media/binary files, and over-scans when a root-level path is indexed
**Version:** leann-core 0.3.7
**Env:** WSL2, Python 3.12

### Problem A — `leann watch` hard-crashes on media/binary files
`leann watch`'s change detection hashes file *contents* via a LlamaIndex `SimpleDirectoryReader` with the full default extractor set:

```
cli.py _detect_build_changes → sync.py detect_changes → sync.py generate_file_hashes
  → reader.iter_data() → SimpleDirectoryReader.load_file → readers/file/video_audio/base.py
  → ImportError: Please install OpenAI whisper ...
```

So a single audio/video file anywhere in the scanned roots takes the whole watcher down with an unhandled `ImportError` (image/binary files also emit load errors). Notably, `leann build` survives the same tree because it honors `--file-types` (`required_exts`), but the watch/sync hash scan does **not** apply that filter — an inconsistency. Change detection should restrict to the index's configured file types (or at least skip unreadable files gracefully) rather than instantiate media readers.

### Problem B — indexing a root-level file expands the scan to the whole repo
If `--docs` includes any loose file at the repo root (e.g. `README.md`), the watcher's resolved sync roots collapse to the repo root, so it then crawls everything — `node_modules/`, build dirs, vendored mirrors — and trips Problem A on whatever binary/media it finds:

```
📂 Indexing 15 paths:
    1. /home/.../my-repo          ← entire repo
Failed to load .../assets/img/icon/wasm.png ...
ImportError: Please install OpenAI whisper ...
```

### Repro
1. Build an index whose `--docs` includes a repo-root file plus subdirs, in a repo that also contains any media/binary file (images count), e.g. `leann build x --docs ./src README.md --file-types .py,.md`.
2. Start `leann watch x` → it scans the repo root and crashes on the first media file.

### Expected
Watch honors the index's configured file types / scan scope, and never crashes on files it wouldn't index anyway.

### Workaround
Pass only clean *subdirectories* to `--docs` (no root-level files), and keep media/binaries out of indexed dirs.

ASuresh0524:
Replace SimpleDirectoryReader in change detection with extension-filtered filesystem walks and raw byte hashing. Store explicit files vs directories in sync_roots.json so a loose README.md no longer watches the whole repo.

Fixes #345 

## Checklist

### DIFF
--- a/packages/leann-core/src/leann/cli.py
+++ b/packages/leann-core/src/leann/cli.py
@@ -28,7 +28,7 @@ from .settings import (
     resolve_openai_api_key,
     resolve_openai_base_url,
 )
-from .sync import FileSynchronizer
+from .sync import DEFAULT_INDEX_EXTENSIONS, FileSynchronizer, parse_include_extensions
 
 
 def _normalize_path(path: str) -> str:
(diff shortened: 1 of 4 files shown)


ASuresh0524:
@ArtifexSystems got a fix up for this — PR should be open shortly. watch no longer runs LlamaIndex extractors for change detection (byte hash + same ext filter as build), and loose files like README.md don't expand to watching the whole repo anymore. would love if you can sanity-check on your 4GB/WSL setup when you get a chance 🙏

yichuan-w:
Can you test on a dict with both txt and bin, and if it is good, lets merge that

ASuresh0524:
@yichuan-w yep — tested a dir with both .txt and .bin in the same folder. only the txt gets hashed, bin is skipped, no crash. added test_mixed_txt_and_bin_directory_skips_bin_without_crash in 5266e72, watch/sync tests all green. good to merge from my side 👍

